Issue #22, opened 2022-07-18
[REQUEST] Disable specify configs if the cpu microarchitecture is amd or intel


AldairbotHerotaku:
like on configure put
```
elif [[ "$_microarchitecture" = "45" ]]; then
    scripts/config --disable CONFIG_GENERIC_CPU
    scripts/config --disable CONFIG_AGP_AMD64 
    scripts/config --disable CONFIG_MICROCODE_AMD
    scripts/config --enable CONFIG_MIVYBRIDGE
```
because you can't put a AMD Opteron/Athlon64 on a intel board

ptr1337:
Hey @AldairbotHerotaku 

Thanks for your request. That would work for the processor - select and the configure script for it.
At the auto detect it would be a bit harder to add.

Does this change anything into compatibility ? If you remove the amd-ucode from your system it wont be loaded anyways.
Regards

AldairbotHerotaku:
> Does this change anything into compatibility ? If you remove the amd-ucode from your system it wont be loaded anyways. Regards

I do not speak specifically about the microcode. I talk about loading, for example, the Conf_x86_AMD_platform_Device or rather known as AMD ACPI2PLATFORM Devices Support or in the Intel case the CONFIG_INTEL_IDLE (Cpuidle Driver for Intel Processors)  that is useless if you have an diferent cpu but it is compiled no matter if you use _localmodcfg =
And the only way to remove it would be manually doing it with _makenconfig in each new version of cachyos

There are no problems with compatibility, but it improves the start time a bit (almost invaluable 3 or 4 seconds)

ptr1337:
> > Does this change anything into compatibility ? If you remove the amd-ucode from your system it wont be loaded anyways. Regards
> 
> I do not speak specifically about the microcode. I talk about loading, for example, the Conf_x86_AMD_platform_Device or rather known as AMD ACPI2PLATFORM Devices Support or in the Intel case the Intel Low Power Subsystem Support that is useless if you have an diferent cpu but it is compiled no matter if you use _localmodcfg = And the only way to remove it would be manually doing it with _makenconfig in each new version of cachyos
> 
> There are no problems with compatibility, but it improves the start time a bit (Almost unappreciable 3 or 4 seconds)

I dont really know how we should do this automatically without doing possible breaking from machines.
If you have a idea, feel free to PR it.

PR #29, opened 2022-09-10
Disable Specify configs for MIVYBRIDGE

NextWork123:
This commit is specified for close a issue #22 this need a test.

### DIFF
--- a/linux-bore/auto-cpu-optimization.sh
+++ b/linux-bore/auto-cpu-optimization.sh
@@ -29,6 +29,10 @@ elif [[ ${MARCH} == "GOLDMONT-PLUS" ]]; then
     MARCH="GOLDMONTPLUS"
 elif [[ ${MARCH} == "SKYLAKE-AVX512" ]]; then
     MARCH="SKYLAKE2"
+elif [[ ${MARCH} == "MIVYBRIDGE" ]]; then
+    scripts/config --disable CONFIG_AGP_AMD64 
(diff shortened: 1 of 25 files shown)


linux-bore/auto-cpu-optimization.sh
```diff
@@ -29,6 +29,10 @@ elif [[ ${MARCH} == "GOLDMONT-PLUS" ]]; then
     MARCH="GOLDMONTPLUS"
 elif [[ ${MARCH} == "SKYLAKE-AVX512" ]]; then
     MARCH="SKYLAKE2"
+elif [[ ${MARCH} == "MIVYBRIDGE" ]]; then
+    scripts/config --disable CONFIG_AGP_AMD64
```
ptr1337, line 33: As we already wrote privately we could probably add this for every INTEL based CPU. 
Or we will ad it at the configure script to MNATIVE_INTEL / MNATIVE_AMD. 

All in all it would be nice if you can test it, since so far I know you have a INTEL CPU

ptr1337:
Otherwise it seems good. We can extend it to other marches. 
Would be great if you would sign the commits, so that I will be able to merge this

ptr1337 (approved):

linux-bore/auto-cpu-optimization.sh
sirlucjan: Seems good

sirlucjan (approved):

PR #35, opened 2022-10-01
Use environment variables to override parameters in PKGBUILD

ventureoo:
As discussed in Discord.

The default value is now set using extended Bash substitution, where
first the value of the environment variable is checked, and if the user
has not set it, the to assign it a default value, which is written in
format:

`_var=${_var-default}`

This allows you to both set another variable value and unset the
variable. For example:

`env _use_auto_optimization= _processor_opt=native_intel makepkg -sricCf`

It is possible to use export as well.

All changes were made through the following command
`find . -name "PKGBUILD" | xargs -I {} sed -i '1,178s/\(^_[A-Za-z_][^\n].*\)\(=.*\)/\1=\$\{\1\2\}/g' {}`

### DIFF
--- a/linux-bore/PKGBUILD
+++ b/linux-bore/PKGBUILD
@@ -7,7 +7,7 @@
 # ATTENTION - one of two predefined values should be selected!
 # 'yes' - enable CachyOS config
 # 'no' - disable CachyOS config
-_cachy_config=${_cachy_config='yes'}
+_cachy_config=${_cachy_config-'yes'}
 
 ### Selecting the CPU scheduler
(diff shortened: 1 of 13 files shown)


ptr1337 (approved):
Just tested it locally. Seems to work fine. Some adjustments needed for the buildscript, but should work easily. 

Thanks for the PR. Waiting for @sirlucjan review

ptr1337:
Could you maybe sign your commits?

sirlucjan (approved):

ptr1337:
Merged via cherry picking. Ill close here.

PR #37, opened 2022-10-07
Update README.md

SoongVilda:
We can extract and enter into the folder faster :D


ptr1337 (approved):
Well, should make no difference. But merged

PR #42, opened 2022-10-16
Update README.md

SoongVilda:
@vnepogodin  I think it shouldn't be written as `etherium.` The correct should be `Ethereum.`


vnepogodin (approved):
Yes. Thanks

PR #45, opened 2022-11-03
Remove NEST

SoongVilda:
Is that still relevant? Based on a discussion on Discord, I don't see a reason to have any more in a list of features.


ptr1337:
> Is that still relevant? Based on a discussion on Discord, I don't see a reason to have any more in a list of features.

Actually its just in as "feature" that we provide a additional option for, when people compile there kernel on there own.
The global enabling attempt of NEST was just a try from mu and me, but it seemd that setting all tasks with NEST enabled, brings a temperature regression.

So as long the original patch is used, and the process is started via `taskset`, then NEST still works good.f
Actually we need to watch, if we get it updated against 6.1, if not it will be removed there anyways.

PR #47, opened 2022-11-18
README: Fix the repository installation command

ventureoo:
... again. Honestly, is it really necessary to install this version of pacman with the repository add?

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ sudo pacman-key --lsign-key F3B607488DB35A47
 
 3. You can download first initial packages
 ```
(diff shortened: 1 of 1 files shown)


sirlucjan (approved):

sirlucjan:
Yes, it is really necessary. Our pacman has a patch from @vnepogodin, so you can install x86_64_v3 packages without problems

ptr1337:
Yes and no.

Architecture = x86_64 x86_64_v3 would be needed in the pacman.conf, otherwise you wont be able to fetch x86_64_v3 packages.

Also if archtitecure=auto is not set, pkgfile will complain

PR #62, opened 2023-02-01
Refactor condition of CPU sched

ventureoo:
Notice:
- hardened is renamed to bore-hardened for clarity (well, there is no such scheduler as "hardened")
- Introduced a local `_die()` function to exit the build

### DIFF
--- a/linux-bore/PKGBUILD
+++ b/linux-bore/PKGBUILD
@@ -18,7 +18,7 @@ _cachy_config=${_cachy_config-'yes'}
 # 'bore' - select 'Burst-Oriented Response Enhancer'
 # 'cfs' - select 'Completely Fair Scheduler'
 # 'tt' - select 'Task Type Scheduler by Hamad Marri'
-# 'hardened' - select 'BORE Scheduler hardened' ## kernel with hardened config and hardening patches with the bore scheduler
+# 'bore-hardened' - select 'BORE Scheduler hardened' ## kernel with hardened config and hardening patches with the bore scheduler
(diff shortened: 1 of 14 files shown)


ptr1337 (approved):
LGTM

sirlucjan (approved):
LGTM

linux-bore/PKGBUILD
```diff
@@ -248,39 +248,22 @@ if [ -n "$_latency_nice" ]; then
...
+    pds|bmq) # BMQ/PDS scheduler
+        source+=("${_patchsource}/sched/0001-prjc-cachy.patch");;
+    tt) ## TT Scheduler
+        source+=("${_patchsource}/sched/0001-tt-cachy.patch");;
+    cacule|cacule-rdb) ## CacULE(-RDB) Scheduler
+        source+=("${_patchsource}/sched/0001-cacULE-cachy.patch");;
+    bore) ## BORE Scheduler with latency_nice
+        [ -n "$_tune_bore" ] && source+=("${_patchsource}/misc/0001-bore-tuning-sysctl.patch")
+        source+=("${_patchsource}/sched/0001-bore-cachy.patch");;
+    bore-hardened) ## Hardened Patches with BORE Scheduler
```
ventureoo, line 262: In original condition, the bore-tuning-sysctl patch was also not applying for hardened. Is that normal?

ptr1337: I need to check if it is applying well, otherwise we can add it

ventureoo: OK.

PR #63, opened 2023-02-01
remove cacULE from linux-cachyos

ptr1337:
I think it is time to drop cacULE from linux-cachyos.
We did hold it very long alive, but from major to major it gets more worse and just "kept alive".
Anyways if people want to use cacULE they can use it for the 5.15 or 6.1 LTS.

Signed-off-by: Peter Jung <admin@ptr1337.dev>


ventureoo (approved):
LGTM

PR #72, opened 2023-02-06
Add a workflow for build and lint

ventureoo:


ptr1337 (approved):
Seems good to me.
This should help us to "catch" issues, specially for zfs.

All good!

ventureoo:
Looks like the build and export of variables is going well. Note that I disabled sum checks (https://github.com/CachyOS/linux-cachyos/pull/72/commits/3b43b5d00e7d094f9fe7678c562422d8a0b0a5d8), because of the enable of `_build_zfs=y` it breaks them. It would be nice to have a separate linux-cachyos-zfs for this.

PR #76, opened 2023-02-10
Refactor build function

ventureoo:
The build was successful, so I think it's OK. (https://github.com/ventureoo/linux-cachyos/actions/runs/4144396315).


sirlucjan (approved):
Very good contribution!

ptr1337 (approved):

PR #77, opened 2023-02-10
Improve _nr_cpus flag

sirlucjan:
The point is that the range is between 2 and 512 [1]. In the current case, if someone selects, for example, 1000, the compilation resets and asks users to select, but does not inform him of the range.
That's why I have cut down the possibility of choosing other than what the kernel gives

1.

```
CONFIG_NR_CPUS_RANGE_BEGIN=2
CONFIG_NR_CPUS_RANGE_END=512
CONFIG_NR_CPUS_DEFAULT=64
CONFIG_NR_CPUS=320
```


ventureoo (approved):
LGTM

ptr1337 (approved):

PR #80, opened 2023-02-20
6.2.0 Release

ptr1337:
Changes:

- Port the config from linux-cachyos-rc to the others
- linux-bore does now enable the "bore-tuning-sysctl" to have a difference from linux-cachyos-bore
- per VMA lock keeps to be disabled. We will provide it with 6.3 as default then
- cacULE Scheduler now fully dropped


sirlucjan (approved):
LGTM

ventureoo (approved):
LGTM

PR #81, opened 2023-02-20
6.1: Add LTS Kernel

ptr1337:
Since 6.1 is now LTS, we will maintain it to have a fallback Kernel with long term support.
The patchset will be a bit stripped then it is right now, but this we will do in future Releases.

The LTS Kernel does use the BORE Scheduler as default. Latency Nice is also enabled to have compatibility with the cachyos ananicy-cpp and its rules.


sirlucjan (approved):
LGTM

ventureoo:
Will this kernel have cacule support?

ptr1337:
> Will this kernel have cacule support?

No, it wont. The cacule patchset for 6.1 has currently a lot of workarounds, reverts and so on that we get it even patched into.
cacULE-RDB does even not compile anymore proper. So I think its not worth to include it there.

PR #94, opened 2023-04-11
Remove manual adding cachyos repository

SoongVilda:
I believe this is better because users could edit `/etc/pacman.conf` incorrectly and end up with a broken system. Manual adding is also useless because we already have a script.

README.md is already pretty long and I want to reduce the amount of information and make our README.md more effective.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,37 +101,22 @@ tar xvf cachyos-repo.tar.xz && cd cachyos-repo
 sudo ./cachyos-repo.sh
 ```
 
-#### Behaviour of script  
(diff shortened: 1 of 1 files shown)


ptr1337:
Personally Im against this. This information is quite important for people which would "not" trust our scripts or anything else and want to handle things manually.

ventureoo:
> Personally Im against this. This information is quite important for people which would "not" trust our scripts or anything else and want to handle things manually.

well, I think people if they don't trust us, then they may not trust the binary packages and build themselves...

ventureoo:
I think we should also change the commands to use get script from wget to git + chmod, for people who are paranoid

SoongVilda:
Okay, then you can close the PR and I'll just send another PR only with some sentence fixes.

PR #99, opened 2023-04-19
Add a warning when installing the PDS/BMQ kernel

ventureoo:
Personally, I've wanted to do this for a long time, because it's pretty tedious to get new issues that have nothing to do with our kernel, but are directly related to PDS. I just want the user to know right away where to direct specific issues, because linux-cachyos is not a PDS targeted kernel unlike linux-tkg.

@sirlucjan  @ptr1337 would like to know your opinion.


linux-bore/PKGBUILD
```diff
@@ -247,7 +247,8 @@ case "$_cpusched" in
     cachyos) # CachyOS Scheduler (EEVDF)
         source+=("${_patchsource}/sched/0001-EEVDF.patch");;
     pds|bmq) # BMQ/PDS scheduler
-        source+=("${_patchsource}/sched/0001-prjc-cachy.patch");;
+        source+=("${_patchsource}/sched/0001-prjc-cachy.patch"
```
sirlucjan, line 250: Maybe we should change linux-cachyos-pds.install to linux-cachyos-prjc.install (for both bmq and pds)?

sirlucjan (changes requested):

linux-bore/linux-cachyos-pds.install
```diff
@@ -0,0 +1,16 @@
+post_install() {
+    cat << EOF
+You install the linux-cachyos kernel variant with the PDS/BMQ scheduler.
+Please note that this scheduler can cause various performance and stability
+issues (especially on AMD Ryzen CPUs) and we don't provide official
+support for fixing them. If you encounter any problems while using the kernel,
+please make sure that that it is not a PDS-specific problem before reporting it
```
sirlucjan, line 7: Also here we should change it to "not a PRJC-specific problem"

ventureoo: I think it is better to use PDS for simplicity, because the PRJC project name is not always obvious to ordinary users. As I see it, "not a BMQ or PDS specific problem" is more suitable

sirlucjan: Yes, it is better than just PDS.

ventureoo: Done.

sirlucjan (approved):
LGTM now.

ptr1337 (approved):
Fine for me. 

LGTM

PR #101, opened 2023-04-24
6.3.0-1

ptr1337:
This updates the PKGBUILDs to 6.3.

Following Changes compared to 6.2:
- per-VMA-locks are as default enabled
- Lenovo Legion Patchset is included
- Surface Patchset is included to provide proper support for it
- NUMA disabled for PDS/BMQ
- Latency Nice for now disabled due compilation issues with 6.3
- ZFS current compiles the module from -git because there is in the staging branches no support for 6.3 right now


ventureoo (approved):
LGTM

PR #104, opened 2023-04-26
Clarify description of the full tickless value

ventureoo:
See https://github.com/Frogging-Family/linux-tkg/pull/636 for more info.

### DIFF
--- a/linux-bore/PKGBUILD
+++ b/linux-bore/PKGBUILD
@@ -74,7 +74,8 @@ _tcp_bbr2=${_tcp_bbr2-y}
 _HZ_ticks=${_HZ_ticks-500}
 
 ## Choose between perodic, idle or full
-### Full tickless can give higher performances in various cases but, depending on hardware, lower consistency.
+### Full tickless can give higher performances in case you use isolation of CPUs for tasks
+### and it works only when using the nohz_full kernel parameter, otherwise behaves like idle.
 _tickrate=${_tickrate-full}
 
 ## Choose between full(low-latency), voluntary or server
--- a/linux-cachyos-bmq/PKGBUILD
+++ b/linux-cachyos-bmq/PKGBUILD
@@ -74,7 +74,8 @@ _tcp_bbr2=${_tcp_bbr2-y}
 _HZ_ticks=${_HZ_ticks-1000}
 
 ## Choose between perodic, idle or full
-### Full tickless can give higher performances in various cases but, depending on hardware, lower consistency.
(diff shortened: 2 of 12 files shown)


sirlucjan (approved):

sirlucjan:
Could you update all PKGBUILDS?

ptr1337:
That is not correct.
In 6.1 (or 6.2 not sure) where RCU_NOCB_DEFAULT_ALL introduced, which does automatically offload CPUs.
The kernel parameter is not required then. We have this option as default since 6.1 enabled.

if nohzfull kernel parameter is set, then it will respect that.

ventureoo:
> That is not correct. In 6.1 (or 6.2 not sure) where RCU_NOCB_DEFAULT_ALL introduced, which does automatically offload CPUs. The kernel parameter is not required then. We have this option as default since 6.1 enabled.
> 
> if nohzfull kernel parameter is set, then it will respect that.

Please read my explanation in PR for TKG. Even if the parameter is not required, it still doesn't work properly, since nohz_full implies that we don't make ticks when we have one task running per-CPU core. This is not possible in desktop tasks when too many processes are fighting for CPU time.

ventureoo:
> Could you update all PKGBUILDS?

I added the missing one for linux-cachyos-server.

ptr1337:
> > That is not correct. In 6.1 (or 6.2 not sure) where RCU_NOCB_DEFAULT_ALL introduced, which does automatically offload CPUs. The kernel parameter is not required then. We have this option as default since 6.1 enabled.
> > if nohzfull kernel parameter is set, then it will respect that.
> 
> Please read my explanation in PR for TKG. Even if the parameter is not required, it still doesn't work properly, since nohz_full implies that we don't make ticks when we have one task running per-CPU core. This is not possible in desktop tasks when too many processes are fighting for CPU time. This is the reason why this parameter is usually paired with isolcpus.

https://lore.kernel.org/lkml/20220422175247.2197949-1-joel@joelfernandes.org/T/

```
On systems with CONFIG_RCU_NOCB_CPU=y, there is no default mask provided
which ends up not offloading any CPU. This patch removes a dependency
from the bootloader having to know about RCU and about how to provide
the mask.

With the new option enabled, all CPUs will be offloaded on boot unless
rcu_nocbs= or rcu_nohz_full= kernel parameters provide a CPU list.
```

ventureoo:
I don't understand how a change in the RCU should generally affect the behavior of the full tickless timer. Anyway, closing.

PR #112, opened 2023-05-28
Fix incorrect behavior BBR(2) with fq_codel

ventureoo:
> Our kernels are defaulting to use FQ_CODEL queueing mechanism and it does not pairs well with BBR2 (our default packet scheduler), even on BBR2 source codes it says that it needs plain FQ with pacing so it works as it should, codel does not support pacing and it brings a lot of packet overhead etc, i was testing plain FQ and BBR2 works a lot better now, even pacman feels faster and steam well it works as it should.

All credit goes to @aarrayy.


ventureoo:
Don't merge for now. I found a better way to do it.

ventureoo:
The commit https://github.com/CachyOS/linux-cachyos/pull/112/commits/67b2f4129f814c2cedd2b140cf488add57e2e734 has a more secure way of not accidentally affecting other network algorithms. It can be merged.

sirlucjan (approved):
LGTM

Issue #114, opened 2023-06-05
Reasons to disable MGLRU in the default config


aruhier:
Hi,
I was surprised to see that Multigen LRU is disabled in the default configs, but enabled by default in the PKGBUILD:
```
# CONFIG_LRU_GEN is not set
# CONFIG_LRU_GEN_ENABLED is not set
```

I use the `linux-cachyos` config on Gentoo, therefore I enabled MGLRU myself, but I was wondering if there was any specific reason for it to be disabled in the config file.

aruhier:
Thanks for the answer!

The base archlinux config seems to enable MG_LRU now, that's why I was surprised: https://gitlab.archlinux.org/archlinux/packaging/packages/linux/-/blob/main/config#L1158

ptr1337:
> Thanks for the answer!
> 
> The base archlinux config seems to enable MG_LRU now, that's why I was surprised: https://gitlab.archlinux.org/archlinux/packaging/packages/linux/-/blob/main/config#L1158

We have updated now the config, I think it should be more clear now :)
Actually we could try to do the config "based" on CachyOS things and make them disabled over the config vars.
Generally that should work already fine.
But for users, which are looking on the config it would be easier I guess.

aruhier:
Thanks a lot!

PR #115, opened 2023-06-05
Move default enabling of MGLRU from PKGBUILD to the config

ventureoo:
MGLRU can be disabled at runtime, so I see no reason why we shouldn't enable it by default in our config, especially when Arch already does.

Fixes: https://github.com/CachyOS/linux-cachyos/issues/114


ventureoo:
> In my opinion, a change in the config is enough. There is no need to change the flag and prevent easy disabling.

Sure. However, I don't see much point in leaving this parameter when it can be disabled right on the fly. Done, anyway.

sirlucjan (approved):
LGTM

PR #123, opened 2023-06-26
Add linux-cachyos-eevdf

sirlucjan:
As @ptr1337 asked - linux-cachyos-eevdf which contains only EEVDF, without BORE


ventureoo (approved):
LGTM

ptr1337 (approved):
Fine! Thank you very much

PR #124, opened 2023-06-27
Use zstd -19 for module compression

sirlucjan:
I prepared for the tests that @ptr1337 wants to run


ventureoo (approved):
If Arch uses it, why not.

vnepogodin (approved):
LGTM

PR #133, opened 2023-07-04
Disable per_vma_lock

sirlucjan:
It seems that per_vma_lock is buggy [1] [2]. Arch Linux has disabled it in its config [3] so I think we would have to do the same. 

1. https://patchwork.kernel.org/project/linux-mm/patch/20230703182150.2193578-1-surenb@google.com/
2. https://bugzilla.kernel.org/show_bug.cgi?id=217624
3. https://gitlab.archlinux.org/archlinux/packaging/packages/linux/-/commit/61f3316308c3adbb7c62f6c6351f965014c4444d


sirlucjan:
pfkernel also disabled per_vma_lock: https://codeberg.org/pf-kernel/linux/commit/c2ee4042aff416ca60cdb8f317a560aa938b7a97

ventureoo (approved):
LGTM. On my laptop, I also encountered a number of obscure hardlocks.

vnepogodin (approved):
LGTM

PR #139, opened 2023-07-17
Reactivate I/O schedulers

aarrayy:
I noticed that for example this line:
_mq_deadline_disable=${_mq_deadline_disable-y}

Wasn't doing anything since mq-deadline gets used anyways, so maybe it's impossible to remove, since this is the case on CachyOS in udev rules, i think it's better to just leave it and let the user decide which I/O scheduler to use at runtime.

Also kyber can be useful in some cases so it's better to have another alternative too.


sirlucjan (approved):
LGTM

ventureoo:
To be honest, I don't really understand why we even need all these options when all I/O schedulers can be disabled right at runtime?

PR #160, opened 2023-08-24
Fix typo in README.md

acopier:
On line 269 instead of Nyx there should be Nix.


ptr1337:
Hi,

Chaotic-AUR did named their repo to "nyx". Reason it is also written like this.

PR #167, opened 2023-09-04
Add support for NVIDIA module build

ventureoo:
Similar to ZFS, this allows us to build an NVIDIA driver module already purpose-built for our kernel. This will allow us to save NVIDIA driver users from constant rebuilding of dkms modules, which is often accompanied by a number of errors.

Note that in PKGBUILD, we will probably need to update ``_nv_ver`` whenever we update the driver itself, so that there is no mismatch in the versions of modules and nvidia-utils in our repository.

Also, note that I did not test the build together with LLVM+LTO, so this may require additional testing/fixes.

### DIFF
--- a/linux-bore/PKGBUILD
+++ b/linux-bore/PKGBUILD
@@ -156,7 +156,8 @@ _use_kcfi=${_use_kcfi-}
 # If you use ZFS, refrain from building the RT kernel
 _build_zfs=${_build_zfs-}
 
-# Build the nvidia module in to the kernel
+# Builds the nvidia module and package it into a own base
+# This does replace the requirement of nvidia-dkms
 _build_nvidia=${_build_nvidia-}
 
 # Enable bcachefs
@@ -814,10 +815,12 @@ _package-nvidia(){
     pkgdesc="nvidia module of ${_nv_ver} driver for the linux-$pkgsuffix kernel"
     depends=("linux-$pkgsuffix=$_kernver" "nvidia-utils=${_nv_ver}" "libglvnd")
     provides=('NVIDIA-MODULE')
+    license=('custom')
 
     cd "${srcdir}/${_nv_pkg}/"
     install -dm755 "$pkgdir/usr/lib/modules/${_kernver}-${pkgsuffix}"
     install -m644 kernel/*.ko "$pkgdir/usr/lib/modules/${_kernver}-${pkgsuffix}"
+    install -Dt "$pkgdir/usr/share/licenses/${pkgname}" -m644 LICENSE
     find "$pkgdir" -name '*.ko' -exec zstd --rm -10 {} +
 }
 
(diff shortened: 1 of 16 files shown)


sirlucjan (approved):
LGTM so far

ptr1337:
Thanks! This seems good.
I would do the next day(s) some locale testing (LTO, THin LTO, together with zfs).

If the testing goes well, we can can build them with 6.5.2 into the repository and could make an announcement in the community, that they test it also.
And over the next months, we will implement it to chwd and the ISO. 

Lets see how it goes! :)

linux-bore/PKGBUILD
```diff
@@ -156,6 +156,9 @@ _use_kcfi=${_use_kcfi-}
 # If you use ZFS, refrain from building the RT kernel
 _build_zfs=${_build_zfs-}
 
+# Build the nvidia module in to the kernel
+_build_nvidia=${_build_nvidia-}
+
```
ptr1337, line 161: I would add here a explanation, what this is exactly doing.
Something like:

```
# Builds the nvidia module and package it into a own base
# This does replace the requirement of nvidia-dkms
```

or something equal

ventureoo: Agreed. Will revise once testing is complete.

ventureoo: Done.

monarc99:
a question as a user:

when you enable _build_nvidia, do you use "updpkgsums" to update the checksums and change PKGBUILD? Or --skipchecksums?

ptr1337:
> a question as a user:
> 
> when you enable _build_nvidia, do you use "updpkgsums" to update the checksums and change PKGBUILD? Or --skipchecksums?

You can use any. When we build the kernel, we have an environment variable to automatically generate it in our script.
But you can just use` --skipchecksums`

ptr1337:
@ventureoo 
Please add also in the build script and CI to build the nvidia module:
https://github.com/CachyOS/linux-cachyos/blob/master/script.sh#L7
https://github.com/CachyOS/linux-cachyos/blob/master/.github/workflows/lint.yml#L31

ventureoo:
> @ventureoo Please add also in the build script and CI to build the nvidia module: https://github.com/CachyOS/linux-cachyos/blob/master/script.sh#L7 https://github.com/CachyOS/linux-cachyos/blob/master/.github/workflows/lint.yml#L31

Done

ptr1337:
> Would it be useful to set _nv_ver=535.104.05 by variable?
> 
> e.g. _nv_ver=${_nv_ver-535.104.05}

Sure ,we can do this.

Hebarch:
Is possible to add Vulkan Dev as an option?

E.g.: _build_nvidia_vulkan_dev=${_build_nvidia_vulkan_dev-y}

i suppose that u need to add this driver branch to your repos, right ?

Thanks in advance :-)

ptr1337:
> Is possible to add Vulkan Dev as an option?
> 
> E.g.: _build_nvidia_vulkan_dev=${_build_nvidia_vulkan_dev-y}
> 
> i suppose that u need to add this driver branch to your repos, right ?
> 
> Thanks in advance :-)

I really think, we should not bloat this to much, since it would also required to change the depend on the nvidia-vulkan-utils.
It shouldn't be to hard, if anyone really wants to use the vulkan driver to change that.

PR #180, opened 2023-10-30
6.6.0-1

ptr1337:
This commit brings the initial changes for 6.6.
Currently there is no other custom scheduler then BORE supported and we need to think about dropping the others in the future - but lets see how it does develop.

sched-ext got added additionally for further testings and as playground kernel.


ventureoo:
> This commit brings the initial changes for 6.6. Currently there is no other custom scheduler then BORE supported and we need to think about dropping the others in the future - but lets see how it does develop.
> 

I guess we could wait on removing other things, but it seems TT should be abandoned now. He's been too dependent on CFS, and it's probably just impossible to rebase it now.

Also, I wanted to take the RT patches out from under ``_cpu_sched`` as a separate option, since we'll probably want to use them with BORE.

linux-cachyos-eevdf/PKGBUILD
```diff
@@ -211,35 +212,35 @@ source=(
...
-        [ -n "$_tune_bore" ] && source+=("${_patchsource}/misc/0001-bore-tuning-sysctl.patch")
-        source+=("${_patchsource}/sched/0001-bore-cachy.patch");;
-    rt) ## CFS with RT patches
+        source+=("${_patchsource}/sched/0001-bore.patch");;
+    rt) ## EEVDF with RT patches
         source+=("${_patchsource}/misc/0001-rt.patch"
                  linux-cachyos-rt.install);;
     hardened) ## Hardened Patches with BORE Scheduler
-        source+=("${_patchsource}/sched/0001-bore-cachy.patch"
+        source+=("${_patchsource}/sched/0001-latency-nice.patch"
```
ventureoo, line 238: Latency nice was just removed by @sirlucjan, so here we have to remove it too.

sirlucjan: I have already reported this to @ptr1337

linux-bore/PKGBUILD
```diff
@@ -822,16 +823,16 @@ _package-nvidia(){
 pkgname=("$pkgbase" "$pkgbase-headers")
 [ -n "$_build_zfs" ] && pkgname+=("$pkgbase-zfs")
 [ -n "$_build_nvidia" ] && pkgname+=("$pkgbase-nvidia")
+[ "$_cpusched" = "sched-ext" ] && pkgname+=("$pkgbase-scheduler")
```
ventureoo, line 826: I think the scheduler suffix might be a little confusing

ptr1337: Maybe adding a "s"?

Actually these are the schedulers compiled for sched-ext

ventureoo: Maybe just add the ``sched-ext-tools`` suffix? I think that would make more sense.

ptr1337: Tools sounds weird in my eyes.
This "kernel" is anyways just a playground to give and bring more support into sched-ext testing and developing. They are searching for more benchmarks and testers with "scx_rusty" and others for gaming, so we want to support there as CachyOS.

I think schedulers sounds good compared to "tools". It basically packages the "schedulers" for sched-ext and not any tools, in my eyes.
Naming is complicated x)

ptr1337:
> > This commit brings the initial changes for 6.6. Currently there is no other custom scheduler then BORE supported and we need to think about dropping the others in the future - but lets see how it does develop.
> 
> I guess we could wait on removing other things, but it seems TT should be abandoned now. He's been too dependent on CFS, and it's probably just impossible to rebase it now.
> 
> Also, I wanted to take the RT patches out from under `_cpu_sched` as a separate option, since we'll probably want to use them with BORE.

Yes, TT is a dead variant. I also don't got any response so far from hamad, if he will work on this in the future.

Sure, you can do this RT wise, but I think offering a OWN kernel (maybe a eevdf-rt-bore variant)

sirlucjan:
@ptr1337 @ventureoo 

I added the missing package (none of us noticed it before), corrected the name and updated the config.

sirlucjan (approved):
I think PKGBUILDs are OK now

PR #182, opened 2023-10-31
Drop PRJC schedulers

ventureoo:
We've been discussing this for a very long time on Discord, but I think now is the time to do it. There were quite a few changes to the scheduler in 6.6, like the introduction of EEVDF. We've also gotten a lot of bugs from using PDS/BMQ for a long time, and our kernel focus has always been on developing and promoting BORE.

A few technical reasons for those people who don't understand why:
- PRJC doesn't support PSI (https://gitlab.com/alfredchen/linux-prjc/-/commit/ceb80e952a426e02d456f5f03b5171649c4e424f). This causes a number of OOM killers that rely on it to break. Waydroid also fails if the kernel was built without PSI support (https://github.com/waydroid/waydroid/issues/136)
- PRJC is having problems building the kernel together with NUMA. Which caused us to have to disable it the whole time. This doesn't break CUDA as the common myth says, but it does make it run slower since CUDA now relies heavily on NUMA.
- Random hangs, some of which are still not fixed:
https://gitlab.com/alfredchen/linux-prjc/-/issues/70
https://gitlab.com/alfredchen/linux-prjc/-/issues/62
- Poor operation on AMD Ryzen CPUs. The largest number of bugs and issues comes as a rule from owners of AMD CPUs.

For those people who still want to use PDS, they can build it either through linux-cachyos-lts, where I still left the option to select PDS. Or revert this commit and build the latest version for 6.5.


ptr1337:
Personally im for this, we have here and there still some prjc users, but these are rare.
I would wait, how alfred will handle the 6.7 release in the next 2-3 weeks otherwise this can be merged.

ventureoo:
Rebased over the latest PKGBUILD changes.

linux-cachyos-bore/PKGBUILD
```diff
@@ -10,9 +10,7 @@
 _cachy_config=${_cachy_config-y}
 
 ### Selecting the CPU scheduler
-# ATTENTION - one of six predefined values should be selected!
-# 'bmq' - select 'BitMap Queue CPU scheduler'
-# 'pds' - select 'Priority and Deadline based Skip list multiple queue CPU scheduler'
+# ATTENTION - one of eight predefined values should be selected!
```
sirlucjan, line 13: Here I think the description needs to be changed (I mean the number eight)

ventureoo: Do you want to replace the comment in general or?

sirlucjan: Yes, let's adapt the description to the current situation.

sirlucjan (approved):
LGTM

PR #183, opened 2023-11-01
Add BORE-RT scheduler

sirlucjan:
Since several schedulers will be unavailable in 6.6, it is worth adding an extra option.


ventureoo (approved):
LGTM

ptr1337:
Fine for me also

PR #198, opened 2023-11-09
6.6 Add scx_layered && set alphabetical order in sched-ext-schedulers package

sirlucjan:
If we will add [sched-ext-latest](https://github.com/CachyOS/kernel-patches/blob/master/6.6/sched/0001-sched-ext-latest.patch) to linux-cachyos, we just need to merge this pull request.


sirlucjan:
We can merge it with 6.6.2

PR #214, opened 2024-01-11
GCC Flags

lseman:
Add suggested ClearLinux Optimization flags.

### DIFF
--- a/linux-cachyos/PKGBUILD
+++ b/linux-cachyos/PKGBUILD
@@ -640,7 +640,11 @@ prepare() {
 
 build() {
     cd ${srcdir}/${_srcname}
-    make ${BUILD_FLAGS[*]} -j$(nproc) all
+    if [ -z "${_use_llvm_lto}" ] && ([ -z "${CC}" ] || [ "${CC}" != "clang" ] || [ "${CC}" == "gcc" ]); then
+        make KCFLAGS="-fivopts -fmodulo-sched" ${BUILD_FLAGS[*]} -j$(nproc) all
+    else
+        make ${BUILD_FLAGS[*]} -j$(nproc) all
+    fi
 
     if [ -n "$_build_nvidia" ]; then
         cd "${srcdir}/${_nv_pkg}/kernel"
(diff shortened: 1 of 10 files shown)


linux-cachyos/PKGBUILD
```diff
@@ -640,7 +640,11 @@ prepare() {
 
 build() {
     cd ${srcdir}/${_srcname}
-    make ${BUILD_FLAGS[*]} -j$(nproc) all
+    if [ -z "${_use_llvm_lto}" ] && ([ -z "${CC}" ] || [ "${CC}" != "clang" ] || [ "${CC}" == "gcc" ]); then
```
ventureoo, line 643: Why do you need all these CC checks? It's enough to check `_use_llvm_lto` to know if we are using Clang or not. You should also use _use_llvm_lto `[ "$_use_llvm_lto" = "none" ]` instead of checking for empty.

ptr1337:
I would actually would do the patching method, e.G if [ "$_use_llvm_lto" = "none" ] --> patch makefile for gcc

ptr1337:
Also you have applied this only the linux-cachyos PKGBUILD, it needs to be applied to all of them

lseman:
About the other PKGBUILDs, I'm waiting if we agree on the flags/the approach before editing them all.

As I replied to ventureoo, my concern with only checking lto would be to overwrite a possible user decision which would like to compile with clang but without lto. But I guess that would not make much sense.

linux-cachyos/PKGBUILD
ventureoo: > My concern for that is to not overwrite user decision. What if he don't want lto, but want to compile with Clang, and has CC=clang in his makepkg?
> 

Then nothing happens. Environment variables like CC and CCX are ignored when building the kernel. You can check this for yourself.  The user should use the only way to enable Clang is through the kernel config.

lseman:
I changed to
```
    if [ "$_use_llvm_lto" = "none" ]; then
        make KCFLAGS="-fivopts -fmodulo-sched" ${BUILD_FLAGS[*]} -j$(nproc) all
    else
        make ${BUILD_FLAGS[*]} -j$(nproc) all
    fi
```

shall we keep this way of apply a .patch just for this?

ptr1337:
This needs to e applied also on the other PKGBUILDs, which are not in the archive directory, besides that fine.

lseman:
> This needs to e applied also on the other PKGBUILDs, which are not in the archive directory, besides that fine.

Which others? I've applied to all PKGBUILDs but the server one.

PR #217, opened 2024-01-20
Fix LTO&kCFI

aarrayy:
This fixes clang from complaining about unsupported flags.


ptr1337 (approved):
LGTM

ventureoo:
linux-cachyos-server was missed there

aarrayy:
I did not include it because it wasn't there even before the fix so i thought i was not meant for the server kernel

ptr1337:
Will be removed anyways, since it gets handled over the Kernels Makefile. Will be in 6.7.2 included.

PR #228, opened 2024-02-22
Update README.md

ShadowOfTheDark669:
The install script is crap and didn't backup my original pacman.conf in the .bak file, instead the .bak file ha the repos present . 

The uninstall script did nothing to remove the keyring and mirrorlist packages nor were the repos removed .

If a user installs the repos manually , then how are they expected to have a .bak file in the manual uninstallation instructions if they weren't told to make one ?

 Almost broke my entire installation becuase of these crap instructions and script .

I propose a simpler and manual installation and removal approach that is free of such system breaking flaws.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,36 +84,6 @@ The repositories contain both Arch Linux and CachyOS packages, which have been r
 
 ## How to add CachyOS repositories
 
-### Option 1: Automated Installation of cachyos repositories
-
-We've made it easy for you! Simply run the following commands to use our helper script that does all the work for you.  😉
-
-Run the following commands:
-1. Get archive with script
-
-```
-wget https://mirror.cachyos.org/cachyos-repo.tar.xz
-```
-> If you don't have `wget`, install it with `sudo pacman -S wget`
-
-2. Extract and enter into the archive
-```
-tar xvf cachyos-repo.tar.xz && cd cachyos-repo
-```
-
-3. Run script with sudo
-```
-sudo ./cachyos-repo.sh
-```
-
-#### Behaviour of script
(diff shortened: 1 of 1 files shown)


vnepogodin:
1) the script does backup your pacman.conf in the bak file (you can clearly see it doing there https://github.com/CachyOS/cachyos-repo-add-script/blob/develop/cachyos-repo.sh#L139).
2) uninstall does remove the repos, and not intended to remove keyring and mirrorlist (there is just no point).
3) manual installation expects you to know what you are doing.
4) that can never break your system.
5) this will never be accepted in the current form and not the place to discuss that

ShadowOfTheDark669:
It didn't do that for me , the .bak backup had the cachyos repositories fully enabled . The repos weren't removed when I ran sudo ./cachyos-repo.sh --remove as clearly after it stated repos removed , the cachyos repositories were synced along with core and extra repositories because the script runs  pacman -Syu after supposedly "removing the repos"

ShadowOfTheDark669:
> 1. the script does backup your pacman.conf in the bak file (you can clearly see it doing there https://github.com/CachyOS/cachyos-repo-add-script/blob/develop/cachyos-repo.sh#L139).
> 2. uninstall does remove the repos, and not intended to remove keyring and mirrorlist (there is just no point).
> 3. manual installation expects you to know what you are doing.
> 4. that can never break your system.
> 5. this will never be accepted in the current form and not the place to discuss that

Also the remove operation backs up the cachyos enabled pacman.conf and overwrites the .bak file backed up earlier for some reason whereas it should restore the .bak file. As here:

`  
 local pacman_conf="/etc/pacman.conf"
    local pacman_conf_cachyos="./pacman.conf"
    local pacman_conf_path_backup="/etc/pacman.conf.bak"


info "CachyOS repo removed"
        mv $pacman_conf_cachyos $pacman_conf 
`

Issue #257, opened 2024-06-10
how to compile with march=x86-64-v3 ?


mabod:
I want to build my own cachyos kernel with `march=x86-64-v3`. How do I do that?

I was not able to figure out how to do that with `_processor_opt` and `_use_auto_optimization`.

ptr1337:
https://github.com/CachyOS/linux-cachyos/blob/master/linux-cachyos/configure#L69

You can use the configure script and need to remove _use_auto_optimization

mabod:
I did change the configure script and hardcoded GENERIC_CPU3. 

But my intention was to solve this by just editing the PKGBUILD. This would be the ideal solution. I do not want to mess with all the other files.

linux-tkg for example offers an option like `_processor_opt="generic_v3"`. This is a much more elegant solution from my point of view.

ptr1337:
Yes, you can do the same here?
Just put it into` _processor_opt` generic_v3 and remove the auto-cpu-optimization.

mabod:
I do this:

```
export _processor_opt="generic_v3"
export _use_auto_optimization=""
export _HZ_ticks="300"
makepkg
```

And it does not work. The config is still with `CONFIG_GENERIC_CPU=y`

mabod:
The PKGBUILD has the following code:

```
    if [ -n "$_processor_opt" ]; then
        MARCH="${_processor_opt^^}"
        MARCH2=M${MARCH}
        scripts/config -k -d CONFIG_GENERIC_CPU
        scripts/config -k -e CONFIG_${MARCH2}
    fi
```

That does not work if I want to configure `march=x86-64-v3` because that needs:

`scripts/config -k -e CONFIG_GENERIC_CPU3`

But MARCH2 introduces an extra character 'M'

mabod:
This is how I do it now:

```
# with -march=x86-64-v3
export _processor_opt="y"
export _use_auto_optimization=""
sed -i 's/MARCH2=.*/MARCH2=GENERIC_CPU3/' PKGBUILD
```

ptr1337:
Yes, we will work on this.

ptr1337:
Could you please retest?

mabod:
It work fine. But setting

```
export _processor_opt="GENERIC_CPU3"
export _use_auto_optimization=""
```
 
I get  `-march=x86-64-v3`

PR #261, opened 2024-06-17
Fix incorrect _processor_opt handling for x86-64 levels

ventureoo:
Fixes: https://github.com/CachyOS/linux-cachyos/issues/257


sirlucjan (approved):
LGTM

PR #262, opened 2024-06-20
Add kernel debug package

ventureoo:
Adds linux-cachyos-debug package with non-stripped vmlinux. This is necessary for us because we unfortunately get reports from users about kernel panics and system hangs, to debug them we need a kernel dump created by kexec+makedumpfile (https://wiki.archlinux.org/title/Kdump), but to debug it we need a non stripped vmlinux. A long-standing issue with the arch package is that it doesn't offer a debug package due to manual strip, see: https://gitlab.archlinux.org/archlinux/packaging/packages/linux/-/issues/2. So we can just make a separate package instead of using the debug option in PKGBUILD, which is desirable to put in our repo.


sirlucjan (approved):
LGTM

ptr1337:
I think we can remove then linux-cachyos-sched-ext-debug.

Will this package be fetchable via debuginfod? If no, could we maybe name it "dbg"? Currently, when we are doing adding packages to the debug repostiory, the *-debug packages gets moved to the debug.cachyos.org repository for debuginfod.

ventureoo:
> I think we can remove then linux-cachyos-sched-ext-debug.
> 
> Will this package be fetchable via debuginfod? If no, could we maybe name it "dbg"? Currently, when we are doing adding packages to the debug repostiory, the *-debug packages gets moved to the debug.cachyos.org repository for debuginfod.

Done. I don't think it can be used by debugfold.

PR #285, opened 2024-08-25
Fix config saving when building inside tmpfs

ventureoo:
When building inside tmpfs, ``$srcdir`` and ``$pkgdir``, their parent directory will not be the directory with PKGBUILD, but /tmp. So we need to read symlink of one of the files in ``$srcdir`` to figure out where the directory with PKGBUILD actually is.


sirlucjan (approved):
LGTM

ptr1337 (approved):
LGTM

PR #288, opened 2024-09-14
Use modprobed-db for CI builds to fix the lack of space

ventureoo:
Unfortunately, I can't find any other way other than to disable “fat” modules like i915/nouveau/amdgpu/network drivers. We don't patch them much though.


ptr1337:
Why not just putting a modprobe.d-db entry in?

ventureoo:
> Why not just putting a modprobe.d-db entry in?

If you already have a prepared module list, then I don't mind.

ptr1337:
> > Why not just putting a modprobe.d-db entry in?
> 
> If you already have a prepared module list, then I don't mind.

https://paste.cachyos.org/p/bbdbe24.txt

Here is mine. We could merge from some other computers too and then use that one

ptr1337 (approved):
LGTM, thank you.

PR #294, opened 2024-09-19
6.11: Update base patch checksums

1Naim:
Since 6.11.1 is going to be released in a week from now, let's do a rebuild to fix setting the "none" ioscheduler.


ptr1337:
You maybe want to push a pkgrel

ptr1337 (approved):

PR #295, opened 2024-09-20
Revert "6.11: Sync config with Arch Linux (#291)"

1Naim:
This reverts commit 25364e9f4e34566a37c4a806c8852e323f7e469c.

Enabling rust support prints out a bunch of objtool warnings. See image below. Furthermore, it seems to break clang+LTO kernels, so this is a change that we absolutely do not want yet.

image


ptr1337 (approved):
Thanks Naim, looks good.

PR #298, opened 2024-09-22
linux-cachyos-echo: Drop

ptr1337:
ECHO Scheduler did not get updated since the 6.9 Kernel.

Since this old Kernel has various security issues we should drop it from the repository as well as dropping the PKGBUILD.

If Hamad considers to maintain ECHO again properly, we can reconsider adding this back.


1Naim (approved):
Really hope Hamad gets around to maintaining it again. On the flipside, Alfred seems to be regularly updating Project C schedulers now so we can consider adding those back to repos.

Otherwise, this PR is warranted, an EOL kernel is very bad to ship.

Xynonners:
ECHO has been updated to 6.12.

ptr1337:
> ECHO has been updated to 6.12.

We will not add it due poor maintainance. This just results into adding and removing it again, latest with the upcoming LTS in December.

PR #299, opened 2024-09-22
linux-cachyos-bmq: Add BMQ Scheduler

ptr1337:
This adds the BMQ Scheduler back to the CachyOS Kernels.

It seems that most issue, why we have removed BMQ has been mostly resolved. PSI gets now also as default disabled to avoid issues with PSI and BMQ.

I do not plan to add currently as default PDS too. This just brings us again compiling more kernels and this requires a lot of time.
Maybe we can add it as a option for the Kernel Manager in the Future.


sirlucjan (approved):

1Naim:
Was there any reason for choosing BMQ in favor of PDS?

ptr1337:
> Was there any reason for choosing BMQ in favor of PDS?

BMQ is generally better for most CPUs then PDS. Also, more actively maintained.

1Naim (approved):

PR #304, opened 2024-10-01
linux-cachyos: Default to ThinLTO

### LINKED ISSUE #286: [RFC] Using Clang (ThinLTO) for the default kernel in the cachyos repository
Hi,

Im considering to use the ThinLTO for the "linux-cachyos" as default in the cachyos repository.
Clang built kernel has been pretty stabilized over the years and is also used by Google in Android as default, together with ThinLTO.

## The Clang built Kernel provides as a bunch of benefits, like:
- Rust support (could be interesting for using https://www.phoronix.com/news/Linux-6.12-DRM-Panic-QR-Code)
- Better Performance --> https://www.phoronix.com/review/clang-lto-kernel

## There are some outstanding issues, which needs to be discussed/fixed:
- ~~nvidia-open-dkms with ThinLTO does not boot, fix can be found here: https://github.com/dell/dkms/pull/417~~ https://github.com/CachyOS/CachyOS-PKGBUILDS/commit/f4020405499a3b0bf5676e9f3994fc3b40c007a8
- ~~Find other dkms modules, which are not compatible~~
- The maintenance of the archlinux clang toolchain
- ~~Test kCFI/FineIBT for the kernel and its impact on performance and modules~~ - Breaks NVIDIA modules

This issue is for general discussions and fiddling out the pending issue, which needs to be resolved.
Also, we should consider providing the default kernel built with GCC too, with a "linux-cachyos-gcc" variant or equal, but this will be only used for the cachyos repository.

## known problematic DKMS Modules:
- ~~vmware-workstation~~n (Maybe we can put this at compilation time, like we do with zfs/nvidia) https://paste.soulharsh007.dev/p/5tY8t6.log - workaround available here: https://github.com/CachyOS/CachyOS-PKGBUILDS/pull/320 can not be distributed, due license
- ~~nvidia-390xx-dkms~~ Fixed by: https://github.com/CachyOS/CachyOS-PKGBUILDS/pull/318
- ~~nvidia-470xx-dkms~~ Fixed by: https://github.com/CachyOS/CachyOS-PKGBUILDS/pull/319
- 8192cu-dkms - https://paste.cachyos.org/p/6059efb.c
- rtl8812au-dkms - https://paste.cachyos.org/p/385e3f2.txt
- rtl8821cu-morrownr-dkms-git - https://paste.cachyos.org/p/7545b73.txt
- ~~rtl88xxau-aircrack-dkms-git - https://paste.cachyos.org/p/2fe7c0c.log~~ - mainly ARM project, ignoring.

## Benchmarking
We should benchmark the ThinLTO kernel compared to the GCC quite much.
There will be a bunch of cachyos-benchmarker tests done, as well as a bigger phoronix test suite.

Aditionally we should test using different marches, e.g v3/znver4 if there are any regressions found

1Naim:
Second and final part of #286. The default kernel will now use clang + ThinLTO as its default configuration.

To accomodate this change, I have added a conditional statement for each `_package` function to replace existing LTO kernels that are already installed to smoothen the transition. Furthermore, I have modified the build script to ensure that all PKGBUILDs will build with GCC first before building the LTO kernels.

We should do a staged rollout for this change to cachyos-testing-v3 -> cachyos-znver4 -> all repos to ensure that everything works correctly (most notably dkms modules, even though they compile successfully, we should ensure complete functionality)

Closes #286

### DIFF
--- a/linux-cachyos/PKGBUILD
+++ b/linux-cachyos/PKGBUILD
@@ -243,7 +243,8 @@ fi
 case "$_cpusched" in
     cachyos) # CachyOS Scheduler (Sched-ext + BORE + Cachy Sauce)
         source+=("${_patchsource}/sched/0001-sched-ext.patch"
-                 "${_patchsource}/sched/0001-bore-cachy-ext.patch");;
+                 "${_patchsource}/sched/0001-bore-cachy-ext.patch"
+                 linux-cachyos.install);;
     bore) ## BORE Scheduler
         source+=("${_patchsource}/sched/0001-bore-cachy.patch");;
     bmq) ## Project C Scheduler
@@ -822,5 +823,7 @@ b2sums=('5720e8dd0bed0247dcfe0bdbce17e36e67bff56811611ae1b04f048b77b6e60c2a48303
         '92e635bb3553896146bea23e5e414853723be9843c94750bc6d4bc327a76925bedb2b3b4481f2816a4691d684e417eae0279c1a2019cb37e9e852914696695da'
         'b1e964389424d43c398a76e7cee16a643ac027722b91fe59022afacb19956db5856b2808ca0dd484f6d0dfc170482982678d7a9a00779d98cd62d5105200a667'
         'cdb2c8d3184c4f0c9c9d50cf78db2e5e4a56eac0162b5d2400277e98c0432fecb96a25c50ef99aae328524bb546c3941e5fb3141e59d4943796574d73cd83182'
+        'c7294a689f70b2a44b0c4e9f00c61dbd59dd7063ecbe18655c4e7f12e21ed7c5bb4f5169f5aa8623b1c59de7b2667facb024913ecb9f4c650dabce4e8a7e5452'
         '019217482184364a3582b567649af9ce9af3f0dc6f6620ea6868ebe011c6e646573e167a574b98515997f829732601f13e8f62e76c8571f05d499d2c96da8976'
-        '13a8f31ba15a007b504f769019530d43b8cb1ec612598940f539ccea842f66ad0981d033a69ef5283c87e4e68abc92ea377e18d0ce930afe441683ebc1f2c99b')
+        '13a8f31ba15a007b504f769019530d43b8cb1ec612598940f539ccea842f66ad0981d033a69ef5283c87e4e68abc92ea377e18d0ce930afe441683ebc1f2c99b'
+        'dbc8885359c814d785f3d2fd316fc3416e1a243ea9a49d2a1a2906c283388a67741f8724cda6897b59483de7bdb671e28fc8b13eb09dd15682a8edbd2389b459')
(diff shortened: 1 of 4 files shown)


1Naim:
Okay so `--printsrcinfo` doesn't seem to honor the if statement inside `_package` but the built package does.

```diff
-pkgname = linux-cachyos
+pkgname = linux-cachyos-gcc
        pkgdesc = The Linux SCHED-EXT + BORE + Cachy Sauce Kernel by CachyOS with other patches and improvements kernel and modules
        depends = coreutils
        depends = kmod
@@ -52,9 +48,9 @@ pkgname = linux-cachyos
        provides = linux-cachyos-lto=6.11.1-2
        replaces = linux-cachyos-lto
```

```sh
❯ pacman -Qi linux-cachyos-gcc
Installed From  : None
Name            : linux-cachyos-gcc
Version         : 6.11.1-2
...
Provides        : VIRTUALBOX-GUEST-MODULES  WIREGUARD-MODULE  KSMBD-MODULE  UKSMD-BUILTIN
...
Required By     : linux-cachyos-gcc-headers
Optional For    : None
Conflicts With  : None
Replaces        : None
```

linux-cachyos/PKGBUILD
```diff
@@ -715,6 +733,12 @@ _package-zfs(){
     provides=('ZFS-MODULE')
     license=('CDDL')
 
+    # Replace LTO kernel with the default kernel
+    if [[ "$_use_llvm_lto" = "thin" || "$_use_llvm_lto" = "full" ]] || [ -n "$_use_kcfi" ]; then
+    provides+=(linux-cachyos-lto-zfs=$_kernver)
+    replaces=(linux-cachyos-lto-zfs)
```
ventureoo, line 739:
```suggestion
        provides+=(linux-cachyos-lto-zfs=$_kernver)
        replaces=(linux-cachyos-lto-zfs)
```

ventureoo:
Is this supposed to be for use on all kernels or just linux-cachyos? I don't see any changes in other PKGBUILDs.

1Naim:
> Is this supposed to be for use on all kernels or just linux-cachyos? I don't see any changes in other PKGBUILDs.

@ptr1337 said only for the default kernel.

linux-cachyos/PKGBUILD
```diff
@@ -619,6 +625,12 @@ _package-headers() {
     pkgdesc="Headers and scripts for building modules for the $pkgdesc kernel"
     depends=('pahole' "${pkgbase}")
 
+    # Replace LTO kernel with the default kernel
+    if [[ "$_use_llvm_lto" = "thin" || "$_use_llvm_lto" = "full" ]] || [ -n "$_use_kcfi" ]; then
```
ventureoo, line 629: Given the number of uses of this condition, it can be put into a separate function so that we can change the condition only once in case of anything. E. g;

```
_is_lto_kernel() {
    [[ "$_use_llvm_lto" = "thin" || "$_use_llvm_lto" = "full" ]] || [ -n "$_use_kcfi" ]
    return $?
}
```

1Naim: Let me know if I did anything wrong in the implementation.

ventureoo: No, it's alright.

ventureoo (approved):
LGTM

1Naim:
Another alternative is to modify it to `use_gcc_suffix` instead. I will make changes with the optimal solution.

ptr1337:
I will push this for now to the Zen4 repository

ptr1337 (approved):
Thanks Naim!

PR #305, opened 2024-10-02
BORE: Restore default base slice value of 2ms

1Naim:
Currently, we set the `min_base_slice_ns` value of BORE to 1000000 (1ms). This was previously added to circumvent some audio crackling and glitches
using 1000Hz kernel. However, a few benchmarks I did (see below) shows that throughput is worse with a base slice of 1ms compared to 2ms.
This behaviour is expected and it doesn't matter much when using BORE as a scheduler because we are not focused on throughput, but rather deskop responsiveness
and interactivity. What was unforeseen however, was the fact that responsiveness under heavy load was also hurt by lowering the base slice to 1ms. 

@firelzrd, the developer of BORE, has also stated that a base slice of anything lower than 2ms will cause excessive scheduling, poor cache efficiency and
increased CPU overhead. It will also lower throughput (as mentioned earlier) which can also harm responsiveness.

Users can compare between these two base slice values by the following command for testing:

```sh
echo 2000000 | sudo tee /sys/kernel/debug/sched/min_base_slice_ns
```

image

### DIFF
--- a/linux-cachyos-bmq/PKGBUILD
+++ b/linux-cachyos-bmq/PKGBUILD
@@ -313,12 +313,12 @@ prepare() {
     [ -z "$_cpusched" ] && _die "The value is empty. Choose the correct one again."
 
     case "$_cpusched" in
-        cachyos) scripts/config -e SCHED_CLASS_EXT -e SCHED_BORE --set-val MIN_BASE_SLICE_NS 1000000;;
-        bore|hardened) scripts/config -e SCHED_BORE --set-val MIN_BASE_SLICE_NS 1000000;;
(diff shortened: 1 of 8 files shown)


ptr1337:
The problem was, that setting the default value together with 1000 Hz introduced audio stutters

1Naim:
> The problem was, that setting the default value together with 1000 Hz introduced audio stutters

Ironically, I get more audio crackles with 1ms than 2ms. Also, the reportee who got the audio stutters was using an absurdly low minimum quant of 32. Setting the base slice to 1ms at runtime also didn't fix it for them which is weird to me.

ptr1337 (approved):

PR #309, opened 2024-10-06
PKGBUILD: Add $use_gcc_suffix

1Naim:
This syncs all PKGBUILDs to the recent change made in https://github.com/CachyOS/linux-cachyos/commit/334ab6e62fa06a83a920070b462dc247a7924882. Functionality is kept the same for all other kernels.

Also fixes a typo made in the aforementioned commit.


ptr1337 (approved):
Thanks Naim!

PR #310, opened 2024-10-06
PKGBUILD: Drop .install script

1Naim:
The .install script we had all this time for the rt kernel was never ran on upgrades and installs due to being "sourced" incorrectly in the PKGBUILD. It should have been "sourced" in `install=` instead for it to be ran. However, sourcing it this way poses a few problems:

1. If we add `install=` along in the case statements for cpusched types, the post install script will be ran for each kernel package installed.
2. We could add it to the _package() function for the kernel, but this needs if statements so that its ran on the correct kernel flavour (clutter IMO). Adding it *only* for the rt kernel in case of `linux-cachyos-rt.install` would make its PKGBUILD inconsistent with others (a case can be made that the default kernel is now inconsistent with the other PKGBUILDs, but the default kernel in nature will always be subject to some changes that will not be seen by the other kernel variants)

Due to the aforementioned problems, drop the post install scripts. This decision was also largely in part of the low usage count of the rt kernel. We will also be removing `linux-cachyos.install` anyway as it is only a temporary addition to smooth out migration.


ptr1337 (approved):

ventureoo:
> the post install script will be ran for each kernel package installed.

Why? I don't get why this has to be called for every kernel. The .install scripts are per-package.

PR #322, opened 2024-11-05
linux-cachyos: AutoFDO Integration

ptr1337:
This is just an example AutoFDO Integration, which users can use.

Currently, I have removed the lto/gcc naming, since AutoFDO only works with Clang.

Following is missing:
- Find a way to integrate AutoFDO usage into all profiles
- Die, if AutoFDO is set, but not clang is used as compiler. Or we enable automatically as ThinLTO and force it
- Propellor Integration: This requires an additonal profiling with the AutoFDO optimized kernel. I dont think, that many users really want to use that, but we might want to integrate this for the default "linux-cachyos" kernel in the future, were we collect several profiles from different machines with the kernel first compiled in the chroot and then to the test machines installed. After that we merge the profiles and use it.


sirlucjan (approved):
It looks good. I'll have time at the end of the week to see how these profiles work.

linux-cachyos-rc/PKGBUILD
```diff
@@ -146,20 +146,27 @@ _build_nvidia_open=${_build_nvidia_open-}
...
+# Enable AUTOFDO_CLANG for the first compilation to create a kernel, which can be used for profiling
+# This only works for Clang compiled kernels, therefore thin or full lto needs to be set in _use_llvm_lto
+# Workflow:
+# https://cachyos.org/blog/2411-kernel-autofdo/
+# 1. Compile Kernel with _autofdo=y
+# 2. Boot the kernel in QEMU or on your system, see Workload
+# 3. Profile the kernel and convert the profile, see Generating the Profile for AutoFDO
+# 4. Put the profile into the sourcedir and name the profile "perf.afdo"
+# 5. Enable _autofdo_use
+_autfodo=${_autofdo-}
```
ventureoo, line 158:
```suggestion
_autofdo=${_autofdo-}
```
Typo

linux-cachyos-rc/PKGBUILD
```diff
@@ -472,6 +479,12 @@ prepare() {
 
     echo "Selecting '$_hugepage' TRANSPARENT_HUGEPAGE config..."
 
+    # Enable Clang AutoFDO
+    # Add additonal check if Thin or Full LTO is enabled otherwise die
+    if [ "$_autofdo" ] || [ -n "$_autofdo_use" ]; then
```
ventureoo, line 484:
```suggestion
    if [ -n "$_autofdo" ] || [ -n "$_autofdo_use" ]; then
```
We need to follow some consistent check style.

linux-cachyos-rc/PKGBUILD
```diff
@@ -556,7 +569,16 @@ prepare() {
 
 build() {
     cd "$_srcname"
-    make "${BUILD_FLAGS[@]}" -j"$(nproc)" all
+
+    # Use AutoFDO Profile named perf.afdo
+    # Maybe we can find a way to make this dyamic, but for now keep it hardcoded
+    # As soon LLVM 19 is in arch repository, we can also think about Propellor
+    if [ -n "$_autofdo_use" ]; then
+        make "${BUILD_FLAGS[@]}" CLANG_AUTOFDO_PROFILE=${srcdir}/perf.afdo -j"$(nproc)" all
```
ventureoo, line 577:
```suggestion
        make "${BUILD_FLAGS[@]}" CLANG_AUTOFDO_PROFILE="${srcdir}"/perf.afdo -j"$(nproc)" all
```

ventureoo: BTW, you can make `CLANG_AUTOFDO_PROFILE` part of `BUILD_FLAGS`, so you don't have to do that check here and you just add a 
```
BUILD_FLAGS+=(CLANG_AUTOFDO_PROFILE="${srcdir}"/perf.afdo)
```
In a general `_autofod` check in `prepare()`.

ptr1337: Thats a good idea

linux-cachyos-rc/PKGBUILD
```diff
@@ -472,6 +486,15 @@ prepare() {
...
     echo "Selecting '$_hugepage' TRANSPARENT_HUGEPAGE config..."
 
+    # Enable Clang AutoFDO
+    # Add additonal check if Thin or Full LTO is enabled otherwise die
+    if [ -n "$_autofdo" ] || [ -n "$_autofdo_use" ]; then
+        scripts/config -e AUTOFDO_CLANG
+    fi
+
+    if [ -n "$_autofdo_use" ]; then
+    BUILD_FLAGS+=(CLANG_AUTOFDO_PROFILE="${srcdir}"/$_autofdo_profile_name)
```
ventureoo, line 496:
```suggestion
        BUILD_FLAGS+=(CLANG_AUTOFDO_PROFILE="${srcdir}"/$_autofdo_profile_name)
```

ptr1337:
According the AutoFDO description in the posted kernel patchset, they are providing following info:

```
One can collect profiles using AutoFDO build for the previous kernel.
AutoFDO employs relative line numbers to match the profiles, offering
some tolerance for source changes. This mode is commonly used in a
production environment for profile collection.
```

This means we can reuse the AutoFDO for like 2-3 minor kernel versions.

Ill create a separate issue for the integration

ptr1337:
Closing in favour of upcoming MR. Keeping that branch for now

PR #323, opened 2024-11-05
Modules: Use all available threads for compressing the modules

ptr1337:
I have noticed today, when compiling the AutoFDO Kernel that the compression was pretty slow.

When adding T0 the compression is speed up and reduces the build time. Since we do not need to care yet about reproducible , this is fine to set


sirlucjan (approved):
LGTM

1Naim (approved):

PR #326, opened 2024-11-11
linux-cachyos-autofdo: Introduce Package

ptr1337:
This kernel is intended to provide a preprofiled kernel, so that user can test this. This will follow the "linux-cachyos" kernel, and after a longer testing period this kernel will be dropped and we are optimizing the linux-cachyos kernel as default.

We will profile the kernel every 2 minor versions with following workloads (as soon LLVM 19 is out):
1. Kernel Compilation
2. cachyos-benchmarker
3. Maybe some other specific workloads, like svt-av1

The profiles needs to be done each architecure. For now we will make the profiles for:
- x86-64-v3
- Zen4

Since there are the most of our users.

Workflow:
1. Compile Kernel with _autofdo=y and _build_debug=y
2. Boot into kenrel in QEMU or host machine
3. Profile the kernel and convert the profile see blogpost
4. Put the profile into the sourcedir andn name the profile perf.afdo
5. Enable _autofdo_use

In the future we will also add propellor


linux-cachyos-autofdo/PKGBUILD
```diff
@@ -0,0 +1,813 @@
...
+
+# Name for the AutoFDO profile
+_autofdo_profile_name=${_autofdo_profile_name-"perf.afdo"}
+
+
+# ATTENTION: Do not modify after this line
+_is_lto_kernel() {
+    [[ "$_use_llvm_lto" = "thin" || "$_use_llvm_lto" = "full" ]] || [ -n "$_use_kcfi" ]
+    return $?
+}
```
1Naim, line 171: We should add a check here if autofdo is used.

linux-cachyos-autofdo/PKGBUILD
```diff
@@ -0,0 +1,813 @@
...
+# Intel CPUs : "mpsc"(P4 & older Netburst based Xeon) "atom" "core2" "nehalem" "westmere" "silvermont" "sandybridge" "ivybridge" "haswell" "broadwell" "skylake" "skylakex" "cannonlake" "icelake" "goldmont" "goldmontplus" "cascadelake" "cooperlake" "tigerlake" "sapphirerapids" "rocketlake" "alderlake"
+# Other options :
+# - "native_amd" (use compiler autodetection - Selecting your arch manually in the list above is recommended instead of this option)
+# - "native_intel" (use compiler autodetection and will prompt for P6_NOPS - Selecting your arch manually in the list above is recommended instead of this option)
+# - "generic" (kernel's default - to share the package between machines with different CPU µarch as long as they are x86-64)
+#
+_processor_opt=${_processor_opt-zen4}
+
+# This does automatically detect your supported CPU and optimizes for it
+_use_auto_optimization=${_use_auto_optimization-}
```
1Naim, line 105:
```suggestion
_processor_opt=${_processor_opt-}

# This does automatically detect your supported CPU and optimizes for it
_use_auto_optimization=${_use_auto_optimization-y}
```

linux-cachyos-autofdo/PKGBUILD
ptr1337: Yes, can be added.

PR #329, opened 2024-11-15
stable: Update to 6.12

1Naim:
6.12 will be released in approx. 3 days, this is the branch that will prepare our PKGBUILDs for that kernel version, as it seems that 6.12 is quite a big release.

Changes in patchset from 6.11 to 6.12:
- Updated AMD V-Cache Driver
- crypto patch that speeds up workloads dealing with crc32 and AES128
- The everchanging fixes patch

Honorable Mentions:
- openvpn-dco - Part of 6.12 patchset until rc7. Module is unused by userspace so there is no reason for us to ship it.

PKGBUILD Changes:
- Updated descriptions and comments for various kernels
- Dropped `linux-cachyos-sched-ext` - sched-ext is upstream now!
- Ported over changes made in `linux-cachyos-rc/PKGBUILD` to the stable kernel PKGBUILDs

Some considerations to be made:
- ~~linux-cachyos-rt-bore: Use PREEMPT_RT instead of PREEMPT_AUTO~~ - We got no choice, PREEMPT_AUTO isn't a thing anymore


linux-cachyos/PKGBUILD
```diff
@@ -215,45 +220,37 @@ fi
...
-                 "${_patchsource}/sched/0001-bore-cachy-ext.patch");;
-    bore) ## BORE Scheduler
+    cachyos|bore) # CachyOS Scheduler (BORE)
         source+=("${_patchsource}/sched/0001-bore-cachy.patch");;
     bmq) ## Project C Scheduler
         source+=("${_patchsource}/sched/0001-prjc-cachy.patch");;
-    eevdf) ## 6.12 EEVDF patches
-        source+=("${_patchsource}/sched/0001-eevdf-next.patch");;
-    rt) ## EEVDF with RT patches
-        source+=("${_patchsource}/misc/0001-rt.patch");;
```
ventureoo, line 248: Why RT patch was removed?

ptr1337: It should be not required anymore, since it got enabled in upstream. If we want to have latest RT features, from the rt tree, then we can keep it, but it also brings more maintenance work

1Naim: The commit of removing RT dependency on CONFIG_EXPERT can also be cherrypicked to, e.g cachy.patch

linux-cachyos-server/PKGBUILD
```diff
@@ -416,7 +412,8 @@ prepare() {
         case "$_preempt" in
             full) scripts/config -e PREEMPT_BUILD -d PREEMPT_NONE -d PREEMPT_VOLUNTARY -e PREEMPT -e PREEMPT_COUNT -e PREEMPTION -e PREEMPT_DYNAMIC;;
             voluntary) scripts/config -e PREEMPT_BUILD -d PREEMPT_NONE -e PREEMPT_VOLUNTARY -d PREEMPT -e PREEMPT_COUNT -e PREEMPTION -d PREEMPT_DYNAMIC;;
-            server) scripts/config -e PREEMPT_NONE_BUILD -e PREEMPT_NONE -d PREEMPT_VOLUNTARY -d PREEMPT -d PREEMPTION -d PREEMPT_DYNAMIC;;
+            none) scripts/config -e PREEMPT_NONE_BUILD -e PREEMPT_NONE -d PREEMPT_VOLUNTARY -d PREEMPT -d PREEMPTION -d PREEMPT_DYNAMIC;;
+            lazy) scripts/config -e PREEMPT_BUILD -d PREEMPT_NONE -d PREEMPT_VOLUNTARY -e PREEMPT_LAZY -e PREEMPT_COUNT -e PREEMPTION -e PREEMPT_DYNAMIC;;
```
1Naim, line 416: We should disable `PREEMPT_DYNAMIC` for servers running `PREEMPT_LAZY` since theres a choice of `PREEMPT_RCU` being made at compile time (can't change with `preempt=` flag).

1Naim: https://lore.kernel.org/lkml/20241106201758.428310-1-ankur.a.arora@oracle.com/ Relevant discussion thread.

linux-cachyos/PKGBUILD
1Naim: https://git.kernel.org/pub/scm/linux/kernel/git/rt/linux-rt-devel.git/log/?h=linux-6.12.y-rt-rebase This is the latest tree of the rt patchset. Initial thoughts are patch 1-3, 6-9 seem nice to have

linux-cachyos-autofdo/PKGBUILD
```diff
@@ -337,9 +332,8 @@ prepare() {
         cachyos|bore|hardened) scripts/config -e SCHED_BORE;;
         bmq) scripts/config -e SCHED_ALT -e SCHED_BMQ;;
         eevdf) ;;
-        rt) scripts/config -e PREEMPT_COUNT -e PREEMPTION -d PREEMPT_VOLUNTARY -d PREEMPT -d PREEMPT_NONE -d PREEMPT_RT -d PREEMPT_DYNAMIC -e PREEMPT_BUILD -e PREEMPT_BUILD_AUTO -e PREEMPT_AUTO;;
-        rt-bore) scripts/config -e SCHED_BORE -e PREEMPT_COUNT -e PREEMPTION -d PREEMPT_VOLUNTARY -d PREEMPT -d PREEMPT_NONE -d PREEMPT_RT -d PREEMPT_DYNAMIC -e PREEMPT_BUILD -e PREEMPT_BUILD_AUTO -e PREEMPT_AUTO;;
-        sched-ext) ;;
+        rt) scripts/config -d PREEMPT -d PREEMPT_DYNAMIC -e PREEMPT_RT;;
```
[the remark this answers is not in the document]
1Naim: Confirmed working. (needed EXPERT for the test but on release we shouldn't need it)

linux-cachyos-server/PKGBUILD
1Naim: Looks like this isn't possible without the above patchset.

linux-cachyos/PKGBUILD
ventureoo: > It should be not required anymore, since it got enabled in upstream. If we want to have latest RT features, from the rt tree, then we can keep it, but it also brings more maintenance work

From the remaining patches, we don't need the ARM/PowerPC/RISCV specific ones too much since we're not targeting them, but PREEMPT_LAZY and i915 patches are also here. The first one we'll want to have anyway for all kernels, but the second one can be separated into a single patch.

https://gitlab.archlinux.org/archlinux/packaging/upstream/linux-rt/-/tree/linux-6.12.y-rt-patches/patches

1Naim: https://github.com/CachyOS/linux/commits/6.12/rt I've cherrypicked some patches from the rt patchset. This seems to meet the standard expectations of being part of our base patchset. The first two commits is also beneficial for normal kernels, and splitting these into different patches makes no sense in my eyes. The other commits are relatively safe to include for non-RT kernels. 

@ptr1337 what do you think?

linux-cachyos/PKGBUILD
```diff
@@ -16,11 +16,10 @@ _cachy_config=${_cachy_config-y}
 # 'bore' - select 'Burst-Oriented Response Enhancer'
 # 'bmq' - select 'BMQ Scheduler'
 # 'hardened' - select 'BORE Scheduler hardened' ## kernel with hardened config and hardening patches with the bore scheduler
-# 'cachyos' - select 'Sched-Ext Scheduler Framework Variant Scheduler with BORE Scheduler'
+# 'cachyos' - select 'CachyOS Default Scheduler (BORE)'
 # 'eevdf' - select 'EEVDF Scheduler'
```
ventureoo, line 20: I'm not sure if we should keep linux-cachyos-eevdf, since it's now essentially EEVDF+sched-ext out of the box.

1Naim: but we dropped linux-cachyos-sched-ext. There's no other kernel that provides using the default EEVDF scheduler as far as I'm aware.

ventureoo: > There's no other kernel that provides using the default EEVDF scheduler as far as I'm aware.

Well, linux-cachyos-server is basically the same as linux-cachyos-eevdf, except for disabling the CachyOS config options.

1Naim: > > There's no other kernel that provides using the default EEVDF scheduler as far as I'm aware.
> 
> Well, linux-cachyos-server is basically the same as linux-cachyos-eevdf, except for disabling the CachyOS config options.

That's not quite true. Aside from having default EEVDF time slices, the server kernel also has a tickrate frequency of 300Hz compared 1000Hz (not ideal for desktop). It is also not a preemptible kernel (pete is currently considering preempt=lazy for server, I don't really like it tho).

ventureoo: > > > There's no other kernel that provides using the default EEVDF scheduler as far as I'm aware.
> > 
> > 
> > Well, linux-cachyos-server is basically the same as linux-cachyos-eevdf, except for disabling the CachyOS config options.
> 
> That's not quite true. Aside from having default EEVDF time slices, the server kernel also has a tickrate frequency of 300Hz compared 1000Hz (not ideal for desktop). It is also not a preemptible kernel (pete is currently considering preempt=lazy for server, I don't really like it tho).

By that I meant “disabling the CachyOS config options”. It's just close to the kernel defaults.

ventureoo: > It is also not a preemptible kernel (pete is currently considering preempt=lazy for server, I don't really like it tho).

The thing is, with PREEMPT_DYNAMIC now, any kernel can be non-preemptable simply by specifying preempt=none in kernel parameters.

1Naim: While that is true, the choice of preempting RCUs or not is decided at compile time and cannot be dynamically changed, see https://lore.kernel.org/lkml/20241106201758.428310-1-ankur.a.arora@oracle.com/T/#m678eb692a0e27e8500a01698e3c3dc82dbbee499. I'm not too sure how significant this is but it is a difference to keep in mind.

ventureoo: > While that is true, the choice of preempting RCUs or not is decided at compile time and cannot be dynamically changed, see https://lore.kernel.org/lkml/20241106201758.428310-1-ankur.a.arora@oracle.com/T/#m678eb692a0e27e8500a01698e3c3dc82dbbee499. I'm not too sure how significant this is but it is a difference to keep in mind.

Well, it makes sense. It's just that in my opinion we have quite a lot of kernels, which makes maintenance difficult and inconsistent between PKGBUILD, as some of them can't be supported on the latest version of kernel, such as it was many times with linux-cachyos-hardened and linux-cachyos-pds (when it was still there), while at the same time most of the users just use linux-cachyos. Like, we now don't have much difference between linux-cachyos-bore and linux-cachyos, since sched-ext is now in both of them. 

Ideally we could just have one PKGBUILD and a system of applying environment variables that would allow build derivatives like linux-cachyos-bmq, linux-cachyos-server and etc, but I haven't figured out how to best design that yet.

1Naim:
`linux-cachyos` and `linux-cachyos-rt-bore` has been compile-tested. Binaries can be found in https://share.cachyos.org/x86_64_v3/kernel/stable-testing/. Patches used are based on https://github.com/CachyOS/linux/commits/6.12/base-dev. I'll try to rebase this PR to fix the merge conflicts once the tarballs are available in https://cdn.kernel.org.

sirlucjan (approved):
It looks OK to me, but with one comment: maybe we would leave the sched-ext kernel and add a patch to it https://git.kernel.org/pub/scm/linux/kernel/git/tj/sched_ext.git/ because I think @htejun  might be interested in that.

ventureoo (changes requested):
```diff
diff --git a/linux-cachyos/PKGBUILD b/linux-cachyos/PKGBUILD
index 3ee2b5ff..6e01b3b3 100644
--- a/linux-cachyos/PKGBUILD
+++ b/linux-cachyos/PKGBUILD
@@ -242,13 +242,12 @@ fi
 
 ## List of CachyOS schedulers
 case "$_cpusched" in
-    cachyos|bore|rt-bore) # CachyOS Scheduler (BORE)
-        source+=("${_patchsource}/sched/0001-bore-cachy.patch");;
+    cachyos|bore|rt-bore|hardened) # CachyOS Scheduler (BORE)
+        source+=("${_patchsource}/sched/0001-bore-cachy.patch");;&
     bmq) ## Project C Scheduler
         source+=("${_patchsource}/sched/0001-prjc-cachy.patch");;
-    hardened) ## Hardened Patches with BORE Scheduler
-        source+=("${_patchsource}/sched/0001-bore-cachy.patch"
-                 "${_patchsource}/misc/0001-hardened.patch");;
+    hardened) ## Hardened Patches
+        source+=("${_patchsource}/misc/0001-hardened.patch");;
 esac
 
 export KBUILD_BUILD_HOST=cachyos
 ```

1Naim:
> CI failed, because .SRCINFO mismatched, but otherwise LGTM.

Yep, was planning on updating .SRCINFO when Pete  synced the patches. All the individual trees should merge fine on a clean 6.12 tree, but IDK which patches are going to be included and which are not, so I'd rather wait.

ptr1337:
Pushed with https://github.com/CachyOS/linux-cachyos/commit/503f4835a29d4ea7925a5fd82344a553915091e1

PR #332, opened 2024-11-20
6.12: Add scx-scheds as optdepend

sirlucjan:
6.12 includes sched-ext, so we need to update dependencies


1Naim (approved):
LGTM, I always wondered why it was a required dependency even though the kernel isn't dependent on it.

On a sidenote, we should probably remove `uksmd` because we have our own ksm stuff via systemd now.

ventureoo (approved):

PR #333, opened 2024-11-20
6.12: Drop uksmd from optdepends

sirlucjan:
Since CachyOS uses https://github.com/CachyOS/CachyOS-Settings/pull/64 by default, we can delete uksmd from optdepends.


1Naim (approved):

ptr1337 (changes requested):
Please generate also .SRCINFO

ventureoo (approved):

ptr1337 (approved):

PR #338, opened 2024-11-27
feat: Use fq packet scheduler when BBR is enabled

1Naim:
BBR relies on pacing that is currently only implemented on fq[1]. While this is not mandatory since [2], research[3][4][5][6] has shown that fq is still preferable to use, not only on BBR but also as a general packet scheduler on host systems. With that in mind, we can adopt using fq as the default globally in our config instead, but since this requires further consideration, let's just set this when BBR is enabled for now.

This is not set on the server kernel where using fq_codel seems to be more desirable.

[1] https://man7.org/linux/man-pages/man8/tc-fq.8.html
[2] https://git.kernel.org/pub/scm/linux/kernel/git/netdev/net-next.git/commit/?id=218af599fa635b107cfe10acf3249c4dfe5e4123
[3] https://research.cec.sc.edu/files/cyberinfra/files/BBR%20-%20Fundamentals%20and%20Updates%202023-08-29.pdf
[4] https://groups.google.com/g/bbr-dev/c/4jL4ropdOV8/m/zNQa-cv7AAAJ - See points 3 and last 2 sentences
[5] https://groups.google.com/g/bbr-dev/c/4jL4ropdOV8/m/0-bNH-KEBgAJ
[6] https://github.com/torvalds/linux/blob/master/net/ipv4/tcp_bbr.c#L55-L57

### DIFF
--- a/linux-cachyos-autofdo/PKGBUILD
+++ b/linux-cachyos-autofdo/PKGBUILD
@@ -454,7 +454,11 @@ prepare() {
             -d DEFAULT_CUBIC \
             -e TCP_CONG_BBR \
             -e DEFAULT_BBR \
-            --set-str DEFAULT_TCP_CONG bbr
+            --set-str DEFAULT_TCP_CONG bbr \
+            -m NET_SCH_FQ_CODEL \
+            -e NET_SCH_FQ \
+            -d CONFIG_DEFAULT_FQ_CODEL \
+            -e CONFIG_DEFAULT_FQ
     fi
 
     ### Select THP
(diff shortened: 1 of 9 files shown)


ventureoo (approved):
LGTM. Honestly, we already had this on by @Shendisx suggestion for BBR2 (https://github.com/CachyOS/linux-cachyos/pull/112), but as I see we lost it when we moved to BBR3 patch.

sirlucjan (approved):
LGTM

PR #353, opened 2024-12-05
fix: Account for autofdo in build script and pkgsuffix conditions

1Naim:
In the PKGBUILD, we have a check whenever $autofdo is a non-zero string, it will build with clang. This breaks
the workflow of the script when building with GCC. It's not that obvious that it's broken because $pkgsuffix is
also broken. This is also fixed by adding a check for it.

We can fix this by forcing $autofdo off during GCC compilations and flipping it back on again when compiling with
ThinLTO.

### DIFF
--- a/linux-cachyos-rc/PKGBUILD
+++ b/linux-cachyos-rc/PKGBUILD
@@ -163,7 +163,7 @@ _is_clang_kernel() {
 
 if [[ "$_use_llvm_lto" = "thin" || "$_use_llvm_lto" = "full" ]] && [ "$_use_lto_suffix" = "y"  ]; then
     _pkgsuffix="cachyos-rc-lto"
-elif [ "$_use_llvm_lto" = "none" ]  && [ -z "$_use_kcfi" ] && [ "$_use_gcc_suffix" = "y" ]; then
+elif [ "$_use_llvm_lto" = "none" ]  && [ -z "$_use_kcfi" ] && [ "$_use_gcc_suffix" = "y" ] && [ -z "$_autofdo" ]; then
     _pkgsuffix="cachyos-rc-gcc"
 else
     _pkgsuffix="cachyos-rc"
--- a/linux-cachyos/PKGBUILD
+++ b/linux-cachyos/PKGBUILD
@@ -163,7 +163,7 @@ _is_clang_kernel() {
 
 if [[ "$_use_llvm_lto" = "thin" || "$_use_llvm_lto" = "full" ]] && [ "$_use_lto_suffix" = "y"  ]; then
     _pkgsuffix=cachyos-lto
-elif [ "$_use_llvm_lto" = "none" ]  && [ -z "$_use_kcfi" ] && [ "$_use_gcc_suffix" = "y" ]; then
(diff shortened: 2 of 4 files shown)


ptr1337 (approved):

linux-cachyos-rc/PKGBUILD
```diff
@@ -163,7 +163,7 @@ _is_clang_kernel() {
 
 if [[ "$_use_llvm_lto" = "thin" || "$_use_llvm_lto" = "full" ]] && [ "$_use_lto_suffix" = "y"  ]; then
     _pkgsuffix="cachyos-rc-lto"
-elif [ "$_use_llvm_lto" = "none" ]  && [ -z "$_use_kcfi" ] && [ "$_use_gcc_suffix" = "y" ]; then
+elif [ "$_use_llvm_lto" = "none" ]  && [ -z "$_use_kcfi" ] && [ "$_use_gcc_suffix" = "y" ] && [ -z "$_autofdo" ]; then
```
ventureoo, line 166:
```suggestion
elif ! _is_clang_kernel && [ "$_use_gcc_suffix" = "y" ]; then
```

linux-cachyos/PKGBUILD
```diff
@@ -163,7 +163,7 @@ _is_clang_kernel() {
 
 if [[ "$_use_llvm_lto" = "thin" || "$_use_llvm_lto" = "full" ]] && [ "$_use_lto_suffix" = "y"  ]; then
     _pkgsuffix=cachyos-lto
-elif [ "$_use_llvm_lto" = "none" ]  && [ -z "$_use_kcfi" ] && [ "$_use_gcc_suffix" = "y" ]; then
+elif [ "$_use_llvm_lto" = "none" ]  && [ -z "$_use_kcfi" ] && [ "$_use_gcc_suffix" = "y" ] && [ -z "$_autofdo" ]; then
```
ventureoo, line 166:
```suggestion
elif ! _is_clang_kernel && [ "$_use_gcc_suffix" = "y" ]; then
```

linux-cachyos-rc/PKGBUILD
ventureoo: The same can be used in the condition above

1Naim: For the first condition it doesn't seem "correct" to mark the kernel as LTO when LTO is not enabled but the other clang features are (kCFI, autofdo). I guess what we can do is make LTO a hard dependency for building with clang to avoid any fallacies. Thoughts? @ptr1337 @ventureoo

ventureoo: > For the first condition it doesn't seem "correct" to mark the kernel as LTO when LTO is not enabled but the other clang features are (kCFI, autofdo).

Hmm... Yeah, I overlooked that a bit, although I'm not sure who has desire to build kernel via Clang without LTO, but it makes sense.

1Naim:
I'll close this and open a new PR. I think hardcoding LTO to clang is the cleaner solution and we don't need any additions to the build scripts for that.

PR #359, opened 2024-12-16
Add support for Propeller Optimization

ptr1337:
This adds support to the PKGBUILD for the Propeller Optimization.
Propeller currently depends on LLVM 19, which should arrive soon.

The workflow would be:
1. Default AutoFDO workflow and enable at the last compilation _propeller to be enabled, so that the config gets enabled
2. Boot into the AutoFDO Kernel with Propeller enabled and profile it
3. Convert the profile with following command and replace prefix with the wanted prefix: create_llvm_prof --binary=/usr/src/debug/linux-cachyos-rc/vmlinux --profile=propeller.data --format=propeller --propeller_output_module_name --out=/home/ptr1337/bench/$prefix_cc_profile.txt --propeller_symorder=/home/ptr1337/bench/$prefix_ld_profile.txt
4. Set the selected prefix in the _propeller_prefix 
5. Compile the final kernel

I did not do enhanced benchmarks, but generally AutoFDO + Propeller should provide an 10% throughput benefit according google. We might want to experiment, which profile workload is suited, and how long/extended the profiling should be

Signed-off-by: Peter Jung <admin@ptr1337.dev>

### DIFF
--- a/linux-cachyos-rc/PKGBUILD
+++ b/linux-cachyos-rc/PKGBUILD
@@ -154,6 +154,18 @@ _autofdo=${_autofdo-}
 # Name for the AutoFDO profile
 _autofdo_profile_name=${_autofdo_profile_name-}
 
+# Propeller should be applied, after the kernel is optimized with AutoFDO
+# Workflow:
+# 1. Proceed with above AutoFDO Optimization, but enable at the final compilation also _propeller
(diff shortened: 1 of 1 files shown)


ptr1337:
Merged with:
https://github.com/CachyOS/linux-cachyos/commit/8dad32c7d17831f08087e52839942c38c1e12df3

Issue #346, opened 2024-12-03
[PKGBUILD] Clang: If the kernel is compiled with clang add clang, llvm, lld to the dependencies of the header package


ptr1337:
Currently, user which are installing the cachyos kernel from the AUR, or chaotic-aur running into issues, because the package does not depend on the llvm toolchain.

Currently, we are only using it for makedepends, if we put full or thin lto.
Since the base kernel, does not need to depend on the llvm toolchain, make the headers depending on llvm toolchain, which is required then for compiling external modules.

1Naim:
```diff
diff --git a/linux-cachyos/PKGBUILD b/linux-cachyos/PKGBUILD
index cfe12603..6f0a6bad 100644
--- a/linux-cachyos/PKGBUILD
+++ b/linux-cachyos/PKGBUILD
@@ -642,6 +642,10 @@ _package-headers() {
     pkgdesc="Headers and scripts for building modules for the $pkgdesc kernel"
     depends=('pahole' "${pkgbase}")
 
+    if _is_clang_kernel; then
+        depends+=(clang llvm lld)
+    fi
+
     cd "${_srcname}"
     local builddir="$pkgdir/usr/lib/modules/$(<version)/build"
```

Something like this should be sufficient, although as seen in #304 previously this breaks .SRCINFO generation when `_is_clang_kernel` is false (this is trivial, the actual package has the correct dependency metadata).

1Naim:
The above isn't a problem for `linux-cachyos` because we already use clang for that for the default. Issues arise for the other kernels which still default to GCC. e.g.:

```diff
diff --git a/linux-cachyos-bore/.SRCINFO b/linux-cachyos-bore/.SRCINFO
index 37c19ca8..ac5294d0 100644
--- a/linux-cachyos-bore/.SRCINFO
+++ b/linux-cachyos-bore/.SRCINFO
@@ -48,3 +48,6 @@ pkgname = linux-cachyos-bore-headers
        pkgdesc = Headers and scripts for building modules for the Linux BORE + Cachy Sauce scheduler Kernel by CachyOS with other patches and improvements kernel
        depends = pahole
        depends = linux-cachyos-bore
+       depends = clang
+       depends = llvm
+       depends = lld
```

ptr1337:
> The above isn't a problem for `linux-cachyos` because we already use clang for that for the default. Issues arise for the other kernels which still default to GCC. e.g.:
> 
> ```diff
[...]

PR #362, opened 2024-12-18
linux-cachyos-hardened: 6.12.5-1

sm9cc:
I noticed that the hardened kernel was outdated but there are patches for 6.12.5.


ptr1337:
We are currently waiting that linux-hardened gets updated in archlinux - I know the patchset exists, but the config is also an important part
https://archlinux.org/packages/extra/x86_64/linux-hardened/

1Naim:
This seems like a very lousy attempt at rebasing the PKGBUILD.
1. The PKGBUILD is lazily copied over from the default kernel without default kernel exclusives:
   a. AutoFDO
   b. Conditional clang depends -- .SRCINFO generation is broken with this and it's been decided that only the default kernel will have this because the default kernel defaults to ThinLTO
   c. Defaulting to ThinLTO 
   d. Using user-generated config

sm9cc:
> This seems like a very lousy attempt at rebasing the PKGBUILD.
Why?, its not cool to to be a cunt, it is better to be clear and constructive, that way I can improve any commits later on.

>     1. The PKGBUILD is lazily copied over from the default kernel without default kernel exclusives:
- Please tell me the none "lazy" way then, why would I make a commit at all if I was trying to be "lazy"?
- All of the kernels PKGBUILD's are based on main PKGBUILD, a simple concept that I followed here.

>        a. AutoFDO
?

>        b. Conditional clang depends -- .SRCINFO generation is broken with this and it's been decided that only the default kernel will have this because the default kernel defaults to ThinLTO
Not sure what you mean, why does .SRCINFO matter, are you submitting this to the aur?

>        c. Defaulting to ThinLTO
Whats wrong with ThinLTO, can you be a little clearer?

>        d. Using user-generated config
Obviously this part is wrong as ptr1337 has already mentioned, once the correct config was available I was going to amend my commit.

---
- The kernel built fine on my end with this PKGBUILD. 
- I thought clang / AutoFDO was the new default for all kernels.

I look forward to hearing your (none cuntish) reply that hopefully clears things up a little more.

Thanks

linux-cachyos-hardened/PKGBUILD
```diff
@@ -106,18 +105,17 @@ _use_auto_optimization=${_use_auto_optimization-y}
 # "full: uses 1 thread for Linking, slow and uses more memory, theoretically with the highest performance gains."
 # "thin: uses multiple threads, faster and uses less memory, may have a lower runtime performance than Full."
 # "none: disable LTO
-_use_llvm_lto=${_use_llvm_lto-none}
+_use_llvm_lto=${_use_llvm_lto-thin}
```
1Naim, line 108: Only the default kernel defaults to ThinLTO, the rest stick to GCC

linux-cachyos-hardened/PKGBUILD
```diff
@@ -106,18 +105,17 @@ _use_auto_optimization=${_use_auto_optimization-y}
...
 # n - disable -lto suffix
 # https://github.com/CachyOS/linux-cachyos/issues/36
-_use_lto_suffix=${_use_lto_suffix-y}
+_use_lto_suffix=${_use_lto_suffix-}
 
 # Use suffix -gcc when requested by the user
-# This was added to facilitate https://github.com/CachyOS/linux-cachyos/issues/286
-_use_gcc_suffix=${_use_gcc_suffix-}
+# Enabled by default to show the difference between LTO kernels and GCC kernels
+_use_gcc_suffix=${_use_gcc_suffix-y}
```
1Naim, line 118: This is only meant to be used by the default kernel to not break the naming scheme for the rest of the kernels.

linux-cachyos-hardened/PKGBUILD
```diff
@@ -143,17 +141,37 @@ _build_nvidia_open=${_build_nvidia_open-}
 # Build a debug package with non-stripped vmlinux
 _build_debug=${_build_debug-}
 
-if [[ "$_use_llvm_lto" = "thin" || "$_use_llvm_lto" = "full" ]] && [ "$_use_lto_suffix" = "y"  ]; then
-     _pkgsuffix="cachyos-${_cpusched}-lto"
-elif [ "$_use_llvm_lto" = "none" ]  && [ -z "$_use_kcfi" ] && [ "$_use_gcc_suffix" = "y" ]; then
-    _pkgsuffix="cachyos-${_cpusched}-gcc"
+# Enable AUTOFDO_CLANG for the first compilation to create a kernel, which can be used for profiling
```
1Naim, line 144: AutoFDO isn't used by any other kernel except for the default and rc so there's not reason to add this.

linux-cachyos-hardened/PKGBUILD
```diff
@@ -143,17 +141,37 @@ _build_nvidia_open=${_build_nvidia_open-}
...
+# ATTENTION: Do not modify after this line
+_is_lto_kernel() {
+    [[ "$_use_llvm_lto" = "thin" || "$_use_llvm_lto" = "full" ]]
+    return $?
+}
+
+if _is_lto_kernel && [ "$_use_lto_suffix" = "y"  ]; then
+    _pkgsuffix=cachyos-hardened-lto
+elif ! _is_lto_kernel && [ "$_use_gcc_suffix" = "y" ]; then
+    _pkgsuffix=cachyos-hardened-gcc
```
1Naim, line 167: Why are you even hardcoding `hardened` here?

linux-cachyos-hardened/PKGBUILD
```diff
@@ -608,12 +636,16 @@ _package-headers() {
     pkgdesc="Headers and scripts for building modules for the $pkgdesc kernel"
     depends=('pahole' "${pkgbase}")
 
+    if _is_lto_kernel; then
+        depends+=(clang llvm lld)
+    fi
```
1Naim, line 641: .SRCINFO doesn't respect any conditionals and will add this regardless. Since the other kernels don't default to LTO, we've decided to not add this for the other PKGBUILDs. Please see https://github.com/CachyOS/linux-cachyos/issues/346.

1Naim (commented):
Here are the reviews if I haven't made myself clear.

1Naim:
> Not sure what you mean, why does .SRCINFO matter, are you submitting this to the aur?

Yes, and `devtools` also utilizies .SRCINFO to grab package metadata.

> Whats wrong with ThinLTO, can you be a little clearer?

Nothing, it's just that the other kernels will default to GCC, and the changes you've made in this PR obviously goes against that decision.

> once the correct config was available I was going to amend my commit

There's a feature to draft pull requests in GitHub. How am I (or any other person other than you) supposed to know if this PR is ready or if there will be any other future changes?

sm9cc:
> > Not sure what you mean, why does .SRCINFO matter, are you submitting this to the aur?
> 
> Yes, and `devtools` also utilizies .SRCINFO to grab package metadata.
> 

Understood.

> > Whats wrong with ThinLTO, can you be a little clearer?
> 
> Nothing, it's just that the other kernels will default to GCC, and the changes you've made in this PR obviously goes against that decision.
> 

I thought clang was the new default for all kernels, excuse me if I am wrong here.

> > once the correct config was available I was going to amend my commit
> 
> There's a feature to draft pull requests in GitHub. How am I (or any other person other than you) supposed to know if this PR is ready or if there will be any other future changes?

You are correct, I should have submitted a draft.

linux-cachyos-hardened/PKGBUILD
sm9cc: Why is this, can you explain? thanks.

linux-cachyos-hardened/PKGBUILD
sm9cc: Whether its used or not is the decision of whoever is building the kernel, also, keeping it there keeps things consistent.

linux-cachyos-hardened/PKGBUILD
sm9cc: Understood, I guess the whole concept here can be improved on.

linux-cachyos-hardened/PKGBUILD
sm9cc: Because it was not tagged with "-hardened" when I built it, so I added this, also, this is the PKGBUILD for the hardened kernel, why does it matter?

linux-cachyos-hardened/PKGBUILD
sm9cc: Understood, although again, I thought all kernels default to LTO / Clang now, if not, would you mind explaining why not? Thanks

linux-cachyos-hardened/PKGBUILD
1Naim: If you want to add it in the name of consistency, then you should add it for the other PKGBUILDs too. The default kernel doesn't need to follow the other PKGBUILDs since it's already different in many ways (ThinLTO default, AutoFDO, more soon)

linux-cachyos-hardened/PKGBUILD
1Naim: It doesn't really matter much because -hardened uses its own configs, but ideally there should only be one PKGBUILD. We have much more in this repo to facilitate the building process to our repositories. Just to add another point, it keeps consistency :) with the rest of the PKGBUILDs that does this.

linux-cachyos-hardened/PKGBUILD
1Naim: https://github.com/CachyOS/linux-cachyos/issues/286. IMO it's to reduce the resources used when building kernels as the other kernel variants aren't used as much if at all compared to the default kernel. Other than that, another mass migration would need to take place by adding provides() replaces() to the PKGBUILDs temporarily which we're not quite fond of doing.

PR #366, opened 2024-12-26
Refactor setting of default values

ventureoo:
It allows to get rid of duplicating the variable name, which helps to avoid typical typo errors and looks prettier.

### DIFF
--- a/linux-cachyos-bmq/PKGBUILD
+++ b/linux-cachyos-bmq/PKGBUILD
@@ -9,7 +9,7 @@
 # Set these variables to ANYTHING that is not null or choose proper variable to enable them
 
 ### Selecting CachyOS config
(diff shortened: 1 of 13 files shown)


ptr1337:
Before pushing this the kernel manager likely needs adjustments.

vnepogodin:
> Before pushing this the kernel manager likely needs adjustments.

we usually adjust kernel manager options after getting pushed into the linux-cachyos repo

sirlucjan (approved):
Tested and everything works OK.

sirlucjan:
Just small nit:

I saw that you rebased PR on top of tree, so you've also update propeller settings.

ptr1337 (approved):

PR #369, opened 2024-12-30
feat $_cpusched Options Cleanup

1Naim:
This PR moves around some of the options and also removes some in an attempt to clean it up and make it more clear to the user.
This will require changes in the kernel-manager, so this should be merged in parallel with #366 so there wouldn't be double work involved.

Notably, fd701cc9c8e7707afddcbced72698d31a4fb415e (`PKGBUILD: Remove cachyos from $cpusched options`) removes the longstanding `cachyos` option that signals the scheduler used in our default kernel. This is such a short-term vision-esqued patch that I'm not really confident about it, and is such a strong contender on getting this dropped from the PR.


sirlucjan (changes requested):
I think that remove "cachyos" sched is a bad idea, sometimes it was useful (with eevdf, with sched-ext)

sirlucjan:
Regarding the reinstatement of RT and PDS - the question is whether someone will undertake the responsibility of managing these changes? PDS itself doesn't raise my objections, I haven't used it for years but I know it has its supporters so as an option alongside BMQ it can be.

1Naim:
> I think that remove "cachyos" sched is a bad idea, sometimes it was useful (with eevdf, with sched-ext)

Right... the idea wasn't well-baked, and I considered removing it from this series too. I think long-term it would benefit us from keeping this option so I'll drop this

linux-cachyos-rc/PKGBUILD
```diff
@@ -329,11 +326,10 @@ prepare() {
     [ -z "$_cpusched" ] && _die "The value is empty. Choose the correct one again."
 
     case "$_cpusched" in
-        cachyos|bore|hardened) scripts/config -e SCHED_BORE;;
+        bore) scripts/config -e SCHED_BORE;;
         bmq) scripts/config -e SCHED_ALT -e SCHED_BMQ;;
+        pds) scripts/config -e SCHED_ALT -e SCHED_PDS;;
         eevdf) ;;
-        rt) scripts/config -d PREEMPT -e PREEMPT_RT;;
```
1Naim, line 335: Right... so doing this is not as simple as moving things to other options. It breaks $pkgsuffix in a way that fixing it would add more complexity than what we have now. Dropping this too...

sirlucjan: Maybe it is better to give up this change too and leave only hardened + add PDS?

sirlucjan (approved):
Tested and everything works OK.

PR #373, opened 2025-01-06
linux-cachyos-rc: Use lazy preemption by default

### LINKED ISSUE #339: (Lazy) Consider to use LAZY Preemption as default on desktop kernels
Lazy Preemption has been recently added to the 6.13 Kernel.
https://lwn.net/Articles/994322/

Generally, Lazy should replace "None" and "Voluntary", while maintaining for RT tasks still the full preemption modell.
We could adjust games in the future to use real time ISO, which allows to use at these full preemption, while other tasks still benefits from using the higher throughput.

Production Benchmarks generally looking good with Lazy Preemption.
Gaming needs to be evaluated.

1Naim:
With promising results shared in https://github.com/CachyOS/linux-cachyos/issues/339, let's enable this and see how
it goes. A prerequisite of this is to disable dynamic preemption, which this PR already does. Further testing can be done with kernels from [1]. 

~The last commit introduces dynamic preempt as an option in the PKGBUILD because it has been disabled by default, but I'm not too sure if we should really add this (hence it's the last commit).~ Dropped and left dynamic preemption enabled for fully preempted kernels.

[1] https://share.cachyos.org/x86_64_v3/kernel/preempt/

Closes #339


1Naim:
https://flightlesssomething.ambrosia.one/benchmark/950 New evidence has shown that lazy preemption is not meant for high CPU load responsiveness. I'm closing this PR as the config changes aren't necessary at all (modifying the config switches are sufficient). I'll consider picking them up again in the 6.13 stable sync.

Note: The 650FPS average of BORE-lazy is misleading. A more accurate average would be around 350.

Issue #341, opened 2024-12-01
gpio_crystalcove fails to initialize


jokubasver:
On Baytrail based tablets, there is an issue with the init order of GPIO/PWM drivers for the Crystal Cove PMIC chip. Crystal Cove has to be initialized before i915 in order for the Crystal Cove to work, which controls the brightness and power of the LCD display.

The error is as follows:
```
i915 0000:00:02.0: cannot find GPIO chip gpio_crystalcove, deferring
i915 0000:00:02.0: [drm] *ERROR* Failed to own gpio for panel control
i915 0000:00:02.0: [drm] *ERROR* [CONNECTOR:86:DSI-1] Failed to get the PMIC PWM chip
```

Check the following bug reports [here](https://gitlab.freedesktop.org/drm/i915/kernel/-/issues/11081) and [here](https://gitlab.freedesktop.org/drm/i915/kernel/-/issues/26).

Since the init order can't be easily changed at the moment, the fix is to make sure that the Crystal Cove GPIO/PWM drivers are built-in, and i915 is built as a module, so that PMIC initializes before i915.

In CachyOS kernel, both Crystal Cove GPIO/PWM and i915 are built as modules.
Please consider changing `CONFIG_GPIO_CRYSTAL_COVE` and `CONFIG_PWM_CRC` from `=m` to `=y` in order to fix this issue.

jokubasver:
Arch kernel has pushed the fix: https://gitlab.archlinux.org/archlinux/packaging/packages/linux/-/commit/e11962a33378d3001d875e4cf3e2a2511199e3b7

ptr1337:
Fixed in https://github.com/CachyOS/linux-cachyos/commit/57a671fd2248be40ea640628e19db3a2eace851f

PR #384, opened 2025-01-23
6.13: Sync config with Arch Linux

1Naim:


linux-cachyos/config
```diff
@@ -10004,7 +9471,7 @@ CONFIG_AMD_PMF=m
 # CONFIG_AMD_PMF_DEBUG is not set
 CONFIG_AMD_PMC=m
 CONFIG_AMD_MP2_STB=y
-CONFIG_AMD_3D_VCACHE=y
+CONFIG_AMD_3D_VCACHE=m
```
1Naim, line 9474: This is a module in Arch Linux, I've synced it here too but we can change it back to be part of the kernel if we want.

linux-cachyos/config
```diff
@@ -5320,14 +5182,11 @@ CONFIG_GPIO_TPIC2810=m
...
 CONFIG_GPIO_ADP5520=m
-CONFIG_GPIO_ADP5585=m
 CONFIG_GPIO_ARIZONA=m
-CONFIG_GPIO_BD71815=m
-CONFIG_GPIO_BD71828=m
 CONFIG_GPIO_BD9571MWV=m
 CONFIG_GPIO_CGBC=m
 CONFIG_GPIO_CROS_EC=m
-CONFIG_GPIO_CRYSTAL_COVE=m
+CONFIG_GPIO_CRYSTAL_COVE=y
```
1Naim, line 5189: https://github.com/CachyOS/linux-cachyos/issues/341

linux-cachyos/config
```diff
@@ -12494,8 +11880,6 @@ CONFIG_ARCH_HAS_DEBUG_VIRTUAL=y
 # CONFIG_DEBUG_VIRTUAL is not set
 CONFIG_DEBUG_MEMORY_INIT=y
 # CONFIG_DEBUG_PER_CPU_MAPS is not set
-CONFIG_ARCH_SUPPORTS_KMAP_LOCAL_FORCE_MAP=y
-# CONFIG_DEBUG_KMAP_LOCAL_FORCE_MAP is not set
```
1Naim, line 12498: So these are exposed if NR_CPUS <= 4096. We're using max now so it's 8192 so I've removed this. Arch doesn't have it too.

linux-cachyos/config
ptr1337: Please change it back to =y, otherwise the user need to modprobe it all the time, if they want to use.

ptr1337:
Outside of the cache driver, looking fine.
Seems like there were quite a bunch of changes, and I just migrated the new ones, but did not remove the old ones.

1Naim:
> Outside of the cache driver, looking fine. Seems like there were quite a bunch of changes, and I just migrated the new ones, but did not remove the old ones.

I think the majority of the diff comes from enabling CONFIG_OF which unlocks these bunch of DRM modules.

linux-cachyos-lts/config
```diff
@@ -128,7 +127,9 @@ CONFIG_BPF_JIT=y
 CONFIG_BPF_JIT_ALWAYS_ON=y
 CONFIG_BPF_JIT_DEFAULT_ON=y
 CONFIG_BPF_UNPRIV_DEFAULT_OFF=y
-# CONFIG_BPF_PRELOAD is not set
+CONFIG_USERMODE_DRIVER=y
+CONFIG_BPF_PRELOAD=y
+CONFIG_BPF_PRELOAD_UMD=m
```
1Naim, line 132: I synced up LTS config too, this is the only meaningful difference.

PR #387, opened 2025-01-28
linux-cachyos-deckify: Re-enable RCU lazy by default

ventureoo:
Perhaps this was just lost along with the last config sync (https://github.com/CachyOS/linux-cachyos/commit/baea61cb7f46030c1f24a65b789370e526629b92), but if it was intentional I'd like to hear reasons, as for Handled devices we'd like to keep battery life more at the expense of small performance regressions.


1Naim (approved):
Indeed

sirlucjan (approved):

ventureoo:
Not sure if we should enable this by default for other kernels as well to help laptop users? I asked about this a long time ago (https://github.com/CachyOS/linux-cachyos/commit/ad001a5708e240e21a8a55785c58dcbd7aa72e1f#commitcomment-143615023), but we decided not to focus on it then. Other kernels and Arch Linux config seem to have this enabled.

1Naim:
I believe having it off by default was a deliberate choice. The wiki has guides on how to enable this via boot parameters, and I've been using that for a long time already.

ptr1337 (approved):

PR #390, opened 2025-01-31
Update README to reflect 6.13 changes

ventureoo:


sirlucjan (approved):

ptr1337:
Maybe should be added that the default kernel is optimized with AutoFDO and Propeller

PR #430, opened 2025-04-07
linux-cachyos-rc: 6.15-rc1

1Naim:
Some changes to PKGBUILD:
- Dropped auto-cpu-optimization.sh
  - The kernel has refactored Makefile CFLAGS additions in 6.15, and the existing patch doesn't work anymore. With the inclusion of CONFIG_X86_NATIVE which uses -march=native from the compiler, I don't really see a need to manually assign each CPU to Kconfig, especially since we don't use it for our own usecase. Therefore, let's remove the script and have users use native instead.
- Do not build NVIDIA and ZFS modules
  - Build is broken, and I've already spent ~6 hours trying to unbreak it, and I can't really go on anymore so I'll try to fix it in the future. In the meantime let's just not build it so that the kernel can still pass building when using the scripts.


TheWyn:
Kernel manager will need an update for dropping the auto cpu optimization

1Naim:
> Kernel manager will need an update for dropping the auto cpu optimization

Yeah I guess so, the options for CPU marches would need to be changed. Alternatively, I can remove the last commit for now so it would just go through and keep it at x86_64v1.

1Naim:
Update: nvidia-open now builds. I didn't notice any performance issues nor any dmesg oddities from brief testing. However, I still consider the last patch (make_device_exclusive()) risky and don't really have the confidence of pushing them.

The patches aren't ready to be shipped in dkms packages either, because some of the last patches still need #ifdefs for compatibility with older kernels. It should be a straightforward process that anyone can do so if anyone is up for that, go for it.

Patches are also randomly added to the closed module, I haven't tested patch applying nor build yet.

1Naim:
> Kernel manager will need an update for dropping the auto cpu optimization

I've dropped https://github.com/CachyOS/linux-cachyos/pull/430/commits/1fb7d1293b5048b19be8b9f6f9905673816743cb so kernel manager won't explicitly fail. It will now just silently fail and fallback to Kconfig defaults.

ptr1337 (approved):
Thanks for the hard NVIDIA work :heart:

PR #436, opened 2025-04-22
linux-cachyos-rc: Add support for Distributed ThinLTO

1Naim:
Add distributed ThinLTO build support for the Linux kernel. This new mode offers several advantages: (1) Increased flexibility in handling user-specified build options. (2) Improved user-friendliness for developers. (3) Greater convenience for integrating with objtool and livepatch.

Note that "distributed" in this context refers to a term that differentiates in-process ThinLTO builds by invoking backend compilation through the linker, not necessarily building in distributed environments.

Depends on dfd4c69f9f12f8fdd68e08cdba250e37cefc93ce so merge this when 6.15rc4 is released

Link: https://lore.kernel.org/20250420010214.1963979-1-xur@google.com/

### DIFF
--- a/linux-cachyos-bmq/PKGBUILD
+++ b/linux-cachyos-bmq/PKGBUILD
@@ -85,7 +85,7 @@
 # ATTENTION - one of three predefined values should be selected!
 # "full: uses 1 thread for Linking, slow and uses more memory, theoretically with the highest performance gains."
 # "thin: uses multiple threads, faster and uses less memory, may have a lower runtime performance than Full."
-# "thin-dist: Similar to thin, but uses a distributed model rather than in-process"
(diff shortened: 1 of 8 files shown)


ptr1337:
I think we should wait till there are a bunch of reviews/acks in upstream before merging this. Otherwise we end again having something in, which will never get pushed, like latency-nice as a good example.

1Naim:
I'm okay with that. Do you still want  to keep the patch around in base or leave it out as an extra patch for now?

ptr1337:
> I'm okay with that. Do you still want to keep the patch around in base or leave it out as an extra patch for now?

I think we can keep it for now for testing reasons.

1Naim:
Well since it looks like it's going to be merged, what say we merge this too?

sirlucjan:
I've rebased PR (I talked with @1Naim about it). thin is still default in linux-cachyos. 
@ptr1337 could you take a look?

Issue #437, opened 2025-04-24
Validity check FAILED (0001-bore-cachy.patch)


KowabungaOfficial:
```
==> WARNING: Skipping verification of source file PGP signatures.
==> Validating source files with b2sums...
    linux-6.14.3.tar.xz ... Passed
    config ... Passed
    auto-cpu-optimization.sh ... Passed
    0001-cachyos-base-all.patch ... Passed
    dkms-clang.patch ... Passed
    0001-bore-cachy.patch ... FAILED
==> ERROR: One or more files did not pass the validity check!
 -> error downloading sources: /home/kwbnga/.cache/yay/linux-cachyos 
         context: exit status 1 
```

This is the upgrade process. Clean or no clean install it does the same thing.

> linux-cachyos-headers 6.14.2-1 -> 6.14.3-1
> linux-cachyos 6.14.2-1 -> 6.14.3-1

- I'm on Arch Linux

PR #438, opened 2025-04-24
chore: Update BORE checksums

1Naim:
The latest handheld patch refresh also updated BORE, even though that wasn't necessary. Nothing changed, so this is fine to merge right away

Fixes: #437


ptr1337:
Can you please keep the pkgrel on 1?

1Naim:
Done.

KowabungaOfficial:
@ptr1337 When will this make it to the aur repo?

PR #454, opened 2025-05-12
PKGBUILD: Enable QR Code Panic for GCC compiled kernels

ptr1337:
This does enable the QR Code Panic Screen for GCC compiled kernels until LLVM (Thin) LTO does support it.

Not applied to all kernels yes, will do this after review

### DIFF
--- a/linux-cachyos/PKGBUILD
+++ b/linux-cachyos/PKGBUILD
@@ -358,6 +358,12 @@ prepare() {
 
     echo "Selecting '$_use_llvm_lto' LLVM level..."
 
+    if [ "$_use_llvm_lto" = "none" ]; then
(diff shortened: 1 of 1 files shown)


ptr1337:
> Any blockers for this? If not I'll add it when syncing 6.15.

Can be put into the 6.15 sync

linux-cachyos/PKGBUILD
```diff
@@ -358,6 +361,12 @@ prepare() {
 
     echo "Selecting '$_use_llvm_lto' LLVM level..."
 
+    if [ "$_use_llvm_lto" = "none" ]; then
```
ventureoo, line 364:
```suggestion
    if ! _is_lto_kernel; then
```

ventureoo (changes requested):

PR #603, opened 2025-11-07
Use extramodules path to install NVIDIA modules

ventureoo:
Since new mkinitcpio added a hook to update initramfs for external modules, we should use this path as well as the nvidia package to get rid of our own hook from cachyos-hooks.


1Naim:
What about ZFS?

ventureoo:
> What about ZFS?

Yep, I forgot about it, probably we should change path for it too.

ventureoo:
> What about ZFS?

Added, although I tested only NVIDIA.

ptr1337 (approved):

PR #614, opened 2025-11-18
6.17: update linux-cachyos-hardened

sirlucjan:
Update to 6.17:

- added missing entry with thin-dist-lto
- added nvidia patch
- sync config


ptr1337 (approved):
Thank you very much!

PR #626, opened 2025-11-26
Fixup for outdated information in the scheduler table

Aarrayy:
BORE is no longer included in our default kernel. I added linux-cachyos to the EEVDF column.


1Naim:
@ptr1337 is the way forward to keep EEVDF as the default choice or are there plans to switch back to BORE, e.g. on 6.18 or something?

ptr1337:
I've been currently a bit split. EEVDF seems to perform quite good currently on the desktop and also we have seen a bit more consistent performance in the phoronix benchmarks

PR #632, opened 2025-11-30
feat: add isp4 patches

darkbasic:
https://www.phoronix.com/news/AMD-ISP4-Driver-v6

This PR adds patches for AMD ISP4 Linux Webcam Driver for HP ZBook Ultra G1a & Future Ryzen Laptops.

Requires https://github.com/CachyOS/kernel-patches/pull/123

### DIFF
--- a/linux-cachyos-bmq/PKGBUILD
+++ b/linux-cachyos-bmq/PKGBUILD
@@ -180,6 +180,13 @@ _nv_open_pkg="NVIDIA-kernel-module-source-${_nv_ver}"
 source=(
     "https://cdn.kernel.org/pub/linux/kernel/v${pkgver%%.*}.x/${_srcname}.tar.xz"
     "config"
+    "${_patchsource}/isp4/0001-amd-isp4.patch"
+    "${_patchsource}/isp4/0002-amd-isp4.patch"
+    "${_patchsource}/isp4/0003-amd-isp4.patch"
+    "${_patchsource}/isp4/0004-amd-isp4.patch"
+    "${_patchsource}/isp4/0005-amd-isp4.patch"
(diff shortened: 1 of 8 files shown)


1Naim:
No, this is not the way. Patchsets must first be pushed in [our kernel tree](https://github.com/CachyOS/linux) and then the kernel-patches repo before making it into this repo.

For patchsets like this, it usually goes into the base-all patchset. We're not accepting any more patches for the 6.17 and 6.18 so this is 6.19 material. I'll pull it if isn't merged in mainline by then.

Thanks. I'll keep the other PR open just so I don't forget.

darkbasic:
So next time I should PR https://github.com/CachyOS/linux directly rather than https://github.com/CachyOS/kernel-patches ?
What's the role of the latter? In the README I read "If you have created a custom patch for the Linux kernel that you would like to contribute to this repository, please fork this repository and submit a pull request with your patch."

1Naim:
> So next time I should PR https://github.com/CachyOS/linux directly rather than https://github.com/CachyOS/kernel-patches ?

The answer isn't as simple as yes or no sadly because patchsets like these usually go into their own branch and you can't merge a PR to a nonexistent branch now can you. The workflow for contributions hasn't been thought out that much yet since its still mostly an internal job.

I think you can submit PRs to either:
1. If you're PRing to the Linux tree then make sure the your branch only contains the patchset you want to submit, and then I'll push that branch to our repo. [Here](https://github.com/CachyOS/linux/commits/6.12/handheld/)'s an example of me picking up @matte-schwartz's branch.
2. if you're PRing to kernel-patches repo then an issue is better because the patchset you're suggesting must be integrated into the base-all patch first and that also means being put into the source tree so it's unlikely that PRs are ever getting accepted here.

PR #634, opened 2025-12-01
6.18.0-1 && 6.12.60-1

ptr1337:
Tested:

Default Kernel with zfs, nvidia and nvidia-open enabled. AutoFDO has been also enabled, profile generation is pending.


ptr1337:
rt patchset is pending. Ive commented it for now

ptr1337:
Ill let it compile tmrw so that testers can fetch it and then expected to land in thursday/friday

PR #636, opened 2025-12-03
feat: Add GPG signature verification for all source tarballs

romeoahmed:
Add GPG signature checks for all upstream Linux kernel source tarballs across all PKGBUILDs.


romeoahmed:
Sorry, but I believe that verifying GPG signatures can prevent **supply chain attacks**, which should benefit every user.
Users who do not use `makepkg` should not be affected.

dougg0k:
I had some doubts on this matter as well, so much that I have asked for the CI's urls before, that ideally should be public / visible to anyone. At least for the kernel it seems to be, with gh actions.

A way to keep integrity is making the whole flow from code to actual release, completely public viewable, and for that, all automated and also public.

So, this is a request that seems very valid to me.

---

I just asked an AI (free chatgpt) to list some additional methods that could still be used even after signing the kernel tar.

- https://pastebin.com/raw/Gz5rnAud
- https://mdbin.sivaramp.com/p/q7dz8du3 better readability

PR #645, opened 2025-12-11
6.18: Prepare for 6.18.1 release

ptr1337:
Just prepartion so i dont need to do on friday


thierryvolpiatto:
Unfortunately, the module r8712u (realtek) is no more present in 6.12 (it was not in 6.18), could you check it in next release?
Thanks.

PR #651, opened 2025-12-17
Include btrfs in kernel instead of as a module

andersrh:
On Fedora the root btrfs filesystem fails to mount, when it's included as a module. Therefore we want to include it in the kernel during build.


1Naim:
This happens when the module isn't in the initramfs. Can you inspect your initramfs if that is the case and maybe fix it in dracut instead?

Additionally I saw that you did https://github.com/CachyOS/copr-linux-cachyos/commit/d8dad26397df66e0dcfce247b382af6ff654eb9a. Instead of bundling a custom config you can add a `scripts/config -e` line on the spec.

andersrh:
Hi @1Naim 

I added it via scripts/config -e in the spec file instead now. I know that a custom config file was a overcomplicated way to do it 😅
I just wanted to make sure to fully replicate the behaviour from the stable kernel where it was set to =y.
And that's also why I wanted to make this PR.

I dont know if there is a specific reason that in LTS it is set to =m and stable set to =y in cachyos?

1Naim:
> I dont know if there is a specific reason that in LTS it is set to =m and stable set to =y in cachyos?

I'm not sure either actually. I guess it was changed at some point? I'll double check in the morning buf if thats the case then theres no reason why this PR shouldn't be merged.

Thanks.

andersrh:
> I'm not sure either actually. I guess it was changed at some point? I'll double check in the morning buf if thats the case then theres no reason why this PR shouldn't be merged.
> 
> Thanks.

okay sounds good 👍 . thank you

1Naim:
It was changed in https://gitlab.archlinux.org/archlinux/packaging/packages/linux/-/merge_requests/8 by Arch and linux-lts config wasn't adapted. Can you port these changes to your PR? Thanks.

1Naim:
Obsolete since LTS is now 6.18

Issue #692, opened 2026-02-07
0001-cachyos-base-all.patch  did not pass the b2sums validity check


x140x1n:
when building manually, not a aur helper caching issue

==> Validating source files with b2sums...
    linux-6.18.9.tar.xz ... Passed
    config ... Passed
    0001-cachyos-base-all.patch ... FAILED
    dkms-clang.patch ... Passed
==> ERROR: One or more files did not pass the validity check!

PR #693, opened 2026-02-08
linux-cachyos: Update checksums

ventureoo:
Closes: https://github.com/CachyOS/linux-cachyos/issues/692


ptr1337:
Github caching catched us again. with 6.19 we gonna switch to git tags

Issue #691, opened 2026-02-06
Move away from kernel-patches repository


ptr1337:
Lets do after 6.19

AndreyGasik:
==> ERROR: The source files failed integrity check!
error: Failed to download sources for "linux-cachyos-6.18.9-1 (linux-cachyos linux-cachyos-headers)":

yes go now

ptr1337:
> ==> ERROR: The source files failed integrity check! error: Failed to download sources for "linux-cachyos-6.18.9-1 (linux-cachyos linux-cachyos-headers)":
> 
> yes go now

Well, this was likely due yay caching ,but in the end it would fix it yeah

AndreyGasik:
/cachyos-repo 56s
❯ sudo ./cachyos-repo.sh
==> Installing CachyOS repo..
gpg: key F3B607488DB35A47: "CachyOS <admin@cachyos.org>" not changed
gpg: Total number processed: 1
gpg:              unchanged: 1
  -> Locally signed 1 key.
==> Updating trust database...
gpg: next trustdb check due at 2026-03-13
error: cachyos-extra-v3: signature from "CachyOS <admin@cachyos.org>" is invalid
error: database 'cachyos-extra-v3' is not valid (invalid or corrupted database (PGP signature))
loading packages...
warning: cachyos-keyring-20240331-1 is up to date -- reinstalling
warning: cachyos-mirrorlist-22-1 is up to date -- reinstalling
warning: cachyos-v3-mirrorlist-22-1 is up to date -- reinstalling
warning: cachyos-v4-mirrorlist-22-1 is up to date -- reinstalling
warning: pacman-7.1.0.r9.g54d9411-2 is up to date -- reinstalling
error: failed to prepare transaction (invalid or corrupted database)

~/cachyos-repo
❯

ptr1337:
As tag scheme we could consider something like that:

```
  ┌──────────────────────────────────┬────────────────────┐
  │          Kernel variant          │    Tag example     │
  ├──────────────────────────────────┼────────────────────┤
  │ Stable initial                   │ cachyos-6.19.0-1   │
  ├──────────────────────────────────┼────────────────────┤
  │ Stable point release             │ cachyos-6.19.3-1   │
  ├──────────────────────────────────┼────────────────────┤
  │ Stable re-spin (patches updated) │ cachyos-6.19.3-2   │
  ├──────────────────────────────────┼────────────────────┤
[...]

PR #700, opened 2026-02-10
PKGBUILD: Switch to git with tags

ptr1337:
This is a draft to switch the PKGBUILD to use git tags for the patch. The current layout would be:

```
  ┌──────────────────────────────────┬────────────────────┐
  │          Kernel variant          │    Tag example     │
  ├──────────────────────────────────┼────────────────────┤
  │ Stable initial                   │ cachyos-6.19.0-1   │
  ├──────────────────────────────────┼────────────────────┤
  │ Stable point release             │ cachyos-6.19.3-1   │
  ├──────────────────────────────────┼────────────────────┤
  │ Stable re-spin (patches updated) │ cachyos-6.19.3-2   │
  ├──────────────────────────────────┼────────────────────┤
  │ LTS                              │ cachyos-6.18.9-1   │
  ├──────────────────────────────────┼────────────────────┤
  │ RC                               │ cachyos-6.19-rc8-1 │
  └──────────────────────────────────┴────────────────────┘
``` 

This would also mean, that when we want to do rebuilds, we need to push a fresh tag. Generally its fine, but we could also introduce an extra variable, which then can be used. Needs to be discussed.

Outside of that the first clone is around 6GB:
```
Receiving objects: 100% (11739472/11739472), 5.85 GiB | 42.50 MiB/s, done.
```

We may can improve the upstream git repository, since it contains a lot of stuff. Otherwise we can also switch away from using "git" and pull .tar.xz from the tags.

The download size, specially for AUR users would be dramatically decreased over the time of building. 

Our build scripts would need to be adjusted to use "SRCDEST", then we can cache the "linux" source.

Fixes: https://github.com/CachyOS/linux-cachyos/issues/691


sirlucjan:
```
diff --git a/PKGBUILD b/PKGBUILD
index 6588e8f..a3aaafc 100644
--- a/PKGBUILD
+++ b/PKGBUILD
@@ -170,16 +170,15 @@ _minor=0
 #_minorc=$((_minor+1))
 #_rcver=rc8
 pkgver=${_major}.${_minor}
-pkgrel=1
 #_stable=${_major}.${_minor}
 _stable=${_major}
 #_stablerc=${_major}-${_rcver}
-_srctag=cachyos-${_major}.${_minor}-${pkgrel}
-_srcname=linux
-#_srcname=linux-${_major}
 pkgdesc='Linux EEVDF + LTO + AutoFDO + Propeller Cachy Sauce Kernel by CachyOS with other patches and improvements.'
+pkgrel=1
 _kernver="$pkgver-$pkgrel"
 _kernuname="${pkgver}-${_pkgsuffix}"
+_srctag=cachyos-${_major}.${_minor}-${pkgrel}
+_srcname=linux-${_srctag}
 arch=('x86_64')
 url="https://github.com/CachyOS/linux-cachyos"
 license=('GPL-2.0-only')
@@ -188,7 +187,6 @@ makedepends=(
   bc
   cpio
   gettext
-  git
   libelf
   pahole
   perl
@@ -206,8 +204,9 @@ _nv_ver=590.48.01
 _nv_pkg="NVIDIA-Linux-x86_64-${_nv_ver}"
 _nv_open_pkg="NVIDIA-kernel-module-source-${_nv_ver}"
 source=(
-    "git+https://github.com/CachyOS/linux.git#tag=${_srctag}"
-    "config")
+    "https://github.com/CachyOS/linux/archive/refs/tags/${_srctag}.tar.gz"
+    "config"
+)
 
 # LLVM makedepends
 if _is_lto_kernel; then
@@ -768,6 +767,6 @@ for _p in "${pkgname[@]}"; do
     }"
 done
 
-b2sums=('9413689f86a18a22d601e8fb019d3a6b1ba4535650dce036cc85368291c10c6adba55655506c4960972b3ca2cc21d054113e280bc6e8efd8a23fbe682ff32918'
+b2sums=('9589df223b0a291852cb405cecdbf248cd936986e77013dee9cdb133c41413250903249b0523e4ae8232b1eb55991a0e36df2d8de14ca40e714b1a5cd16e222f'
         '823b738621a81fa74ed94795004c19ca9e3a17c77ba18f7d83ac3801c2688725c3d8576c83cee99d42e178b95f54c8b4932b321cb15c8209ea965d914600b488'
         'ea26c88950fc06b6ffab93b30e3beacc7d26571a70262334ca8b001dc7899bf96b47d703fbaa7f4e47765c3dafccc23c58a4d4da2169b8ee50012afcb7a1dd96')
-- 
2.53.0
```

```
❯ ls -lah
razem 247M
drwxr-xr-x 2 lucjan lucjan  138 02-10 20:50 .
drwxr-xr-x 6 lucjan lucjan   94 02-10 20:46 ..
-rw-r--r-- 1 lucjan lucjan 246M 02-10 20:46 cachyos-6.19.0-1.tar.gz
-rw-r--r-- 1 lucjan lucjan 287K 02-10 19:30 config
-rw-r--r-- 1 lucjan lucjan 285K 02-10 20:49 config-6.19.0-1-cachyos
-rw-r--r-- 1 lucjan lucjan 1,9K 02-10 20:40 dkms-clang.patch
-rw-r--r-- 1 lucjan lucjan  28K 02-10 20:48 PKGBUILD
-rw-r--r-- 1 lucjan lucjan 2,7K 02-10 19:30 .SRCINFO
```

~250M instead of 5GB

ptr1337:
Yeah, I had this too first but generally I think for AUR it would be better to use git, but well see. Both should work fine

sirlucjan (approved):
Can't wait

ptr1337:
Added example script, which can be used to tag and sign releases when being in cachyos/linux repository. It uses current head.

Input/output looks like this:

```
./tag-release.sh 6.17.13 2

Tag:    cachyos-6.17.13-2
Branch: 6.17/base
Commit: 32d533c69cc2 fixes

Create signed tag? [y/N]
```

A help text is also available with ./tag-release --help.

ptr1337:
Download seems to bit slower:

100 246.0M   0 246.0M   0      0  6.72M      0           00:36          7.43M

Maybe we can just start copying for the default kernels then the tar to all dirs to generate checksums to avoid redownloading.

ptr1337:
> I think you need to separate patchrel from pkgrel, because of >v1 repo manual pkgrel bump.

We generally should not need this anymore after the latest ISO release. Currrently just kept till the next ISO release :)

1Naim:
> > I think you need to separate patchrel from pkgrel, because of >v1 repo manual pkgrel bump.
> 
> We generally should not need this anymore after the latest ISO release. Currrently just kept till the next ISO release :)

What about toolchain/nvidia rebuilds?

ptr1337:
> > > I think you need to separate patchrel from pkgrel, because of >v1 repo manual pkgrel bump.
> > 
> > 
> > We generally should not need this anymore after the latest ISO release. Currrently just kept till the next ISO release :)
> 
> What about toolchain/nvidia rebuilds?

There we could just do a new tag

ptr1337:
> LGTM! This might be a minor inconvenience when bumping hardened and lts kernels.

Indeed, yes. We can look how the general maintenance goes like this and then consider adding a extra variable :)

ptr1337:
Guess we have now the final form. I have changed that we upload the git archive as release, whcih resulted into a 10-15x faster download of the .tar.gz since the fastly cdn from github is used then. This massively improved the download speed.

I have added to the tag-release.sh script to automatically push this when tagging. We **maybe** want to add to the tag-release.sh script following:

- Sign the release with gpg key
- Check, when doing the tag, if we are in the correct  major version (e.g we done some lts pushing, but then tag stable but stayed in lts --> release would use lts and is wrong kernel lol)
- Maybe make it automated.

sirlucjan:
Let's go! 

Yes, later we may add gpg sign. It's good idea but... will our builder have a problem with that?

PR #713, opened 2026-02-17
r8125: use pkgname-based modprobe.d filename to avoid conflicts

ptr1337:
Avoid conflicts on multiple kernels


1Naim:
I still think something like a modprobe config to hijack loading `r8169` is better here. This blacklist just isn't compatible with multiple kernels, especially Arch's. Say you have r8125 installed for default kernel but not for LTS. Booting LTS will render the system with no ethernet drivers at all. Arch's kernel doesn't even have the module so the situation is worse there.

PR #723, opened 2026-02-21
issue_template: Add issue templates for linux-cachyos

ptr1337:


sirlucjan:
Definitely we need something like that

ptr1337:
I will also add that automatically sir_lucjan, naim, ventueroo and me get assigned.

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,157 @@
...
+        - linux-cachyos-rt-bore
+        - linux-cachyos-server
+        - linux-cachyos-deckify
+        - linux-cachyos-rc
+        - linux-cachyos-lto (Clang + ThinLTO)
+        - Other (specify in description)
+    validations:
+      required: true
+
+  - type: input
```
1Naim: could probably replace all this with cachyos-bugreport.sh output and fallback to manual if not possible.

1Naim (changes requested):
Only checked bug report, others pending.

PR #739, opened 2026-02-28
PKGBUILD: add "clang" mode and rename _use_llvm_lto to _use_llvm

ptr1337:
Add a new "clang" option to _use_llvm (formerly _use_llvm_lto) that uses Clang as the compiler with LTO disabled. This is useful for AutoFDO/Propeller profiling workflows and testing without the overhead of LTO linking.

I will run some benchmarks to test the performance between ThinLTO + Propeller and Clang + Propeller. The main benefit we would have here, is that we can enable support for Rust and preparing for the future in case more and more drivers get depending on it.

The QR_CODE Bluescreen is pretty nice and we can also implement an own site for this.
Will enhance to other PKGBUILDs when reviewed

### DIFF
--- a/linux-cachyos/PKGBUILD
+++ b/linux-cachyos/PKGBUILD
@@ -370,7 +370,7 @@ prepare() {
 
     echo "Selecting '$_use_llvm' LLVM level..."
 
-    if ! _is_lto_kernel; then
+    if [[ "$_use_llvm" = "none" || "$_use_llvm" = "clang" ]]; then
         echo "Enabling QR Code Panic for non-LTO Kernels"
         scripts/config --set-str DRM_PANIC_SCREEN qr_code -e DRM_PANIC_SCREEN_QR_CODE \
             --set-str DRM_PANIC_SCREEN_QR_CODE_URL https://panic.archlinux.org/panic_report# \
--- a/linux-cachyos/PKGBUILD
+++ b/linux-cachyos/PKGBUILD
@@ -86,13 +86,14 @@
 # - "generic" (kernel's default - to share the package between machines with different CPU µarch as long as they are x86-64)
 : "${_processor_opt:=}"
 
-# Clang LTO mode, only available with the "llvm" compiler - options are "none", "full" or "thin".
-# ATTENTION - one of three predefined values should be selected!
-# "full: uses 1 thread for Linking, slow and uses more memory, theoretically with the highest performance gains."
-# "thin: uses multiple threads, faster and uses less memory, may have a lower runtime performance than Full."
-# "thin-dist: Similar to thin, but uses a distributed model rather than in-process: https://discourse.llvm.org/t/rfc-distributed-thinlto-build-for-kernel/85934"
-# "none: disable LTO
-: "${_use_llvm_lto:=thin}"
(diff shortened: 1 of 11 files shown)


linux-cachyos/PKGBUILD
```diff
@@ -151,7 +152,12 @@
 
 # ATTENTION: Do not modify after this line
 _is_lto_kernel() {
-    [[ "$_use_llvm_lto" = "thin" || "$_use_llvm_lto" = "full"  || "$_use_llvm_lto" = "thin-dist" ]]
+    [[ "$_use_llvm" = "thin" || "$_use_llvm" = "full"  || "$_use_llvm" = "thin-dist" ]]
+    return $?
+}
+
+_is_clang_build() {
```
ptr1337: Mainly added this for the QR Code check

linux-cachyos/PKGBUILD
```diff
@@ -151,7 +152,12 @@
...
+    _is_lto_kernel || [[ "$_use_llvm" = "clang" ]]
```
1Naim:
```suggestion
_is_clang_build() {
    [[ "$_use_llvm" != "none" ]]
}

is_lto_kernel {
    is_clang_build && [[ "$_use_llvm" != "clang" ]]
}
```

Just a slight nitpick. Either works.

1Naim (commented):
Some comments, I think we should keep _use_llvm_lto as a compat option for something like the kernel manager.

linux-cachyos/PKGBUILD
```diff
@@ -626,6 +632,8 @@ _package-headers() {
     if _is_lto_kernel; then
```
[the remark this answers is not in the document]
1Naim: It wasn't for the user. It was for repo migrations.

ptr1337:
Happy in the discord made some benchmarks with "CachyMod". Generally we can see there:

- Clang only is the slowest
- Clang + Autofdo mid
 -Clang + AutoFDO + ThinLTO --> fastest
 
 This reflects my expections. The difference, outside in the blender benchmark compared to ThinLTO seems ok. So, in case we will ever need Rust support this could be done with disabling ThinLTO
 
 
image

CherrieTheShifter:
Yes This

ptr1337:
@1Naim I think we should proceed with this for 7.2 eventually?

1Naim:
> @1Naim I think we should proceed with this for 7.2 eventually?

yeah, i agree

ptr1337:
@1Naim @ventureoo 

All ready to review. PKGBUILDs are synced, default kernel uses now "Clang".

I gonna write tomorrow to Google and ask what their plans are to support Rust + ThinLTO in the near future.

Scripts have been updated too. We should merge this for 7.2 release then. I think we could theortically exclude the LTS kernel ,but well

linux-cachyos/PKGBUILD
ventureoo: I don't think we should make it like that, since we don't validate user input, so it will be true even if there an invalid value.

ventureoo:
Could we maybe land https://github.com/CachyOS/linux-cachyos/pull/944 first, and then rebase this over it?

CherrieTheShifter:
peak glad this will be done for in 7.2
hope same goes for all prs and such

PR #760, opened 2026-03-10
nvidia: Update to 595.44.02

ptr1337:
We are considering to push the 595 Beta release to the stable repository. I've made a poll in the Discord Thread about the 595 Testing.

Testers reporting with 75% having a "Way better experience then 590", this likely includes the improvements with the sleep behavior, fixes in many games (Monster Hunter Vertex explosion,...) and Wayland HDR improvements.
25% reporting that they have an equal experience as 590.

The Heap vulkan extension wont provide any benefit yet, due not being merged in vkd3d-proton.

Outside of this the new driver also introduces Present_Timing, colorspace_swapchain and HDR_Metadata improving the experience massively.

### DIFF
--- a/linux-cachyos-bmq/PKGBUILD
+++ b/linux-cachyos-bmq/PKGBUILD
@@ -180,7 +180,7 @@ makedepends=(
 )
 
 _patchsource="https://raw.githubusercontent.com/cachyos/kernel-patches/master/${_major}"
-_nv_ver=595.44.02
+_nv_ver=595.45.04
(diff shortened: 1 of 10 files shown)


1Naim:
Why doesn't this match nvidia-utils version?

ptr1337:
> Why doesn't this match nvidia-utils version?

Ups, i likely used nvidia-smi and used the vulkan-beta version x) Thanks!

PR #761, opened 2026-03-13
config: enable pstore debug frontends

Lyapsus:
Body:
## Summary
- enable `CONFIG_PSTORE_CONSOLE`, `CONFIG_PSTORE_PMSG`, and `CONFIG_PSTORE_FTRACE` across all shipped kernel variant configs
- keep this as build-time support only; users still need to opt in by configuring `ramoops` or enabling persistent tracing themselves
- improve post-mortem coverage for hard hangs/watchdog resets without turning on reserved RAM or active tracing by default
## Why
CachyOS already ships `CONFIG_PSTORE`, `CONFIG_PSTORE_RAM`, `CONFIG_FTRACE`, and `CONFIG_DYNAMIC_FTRACE`, but without these frontends a correctly configured `ramoops` setup can still miss useful post-mortem data for some hard hangs.
This change does not request default-on tracing or default reserved RAM, so normal users should not see meaningful runtime impact unless they explicitly configure and use these features.


CherrieTheShifter:
Yes This

PR #768, opened 2026-03-18
Automate patching for nvidia kernel module like in the kernel

1Naim:
This PR splits nvidia patches to its own array that makes it possible to automate patching the kernel module instead of manually adding the patch line.

### DIFF
--- a/linux-cachyos-rc/PKGBUILD
+++ b/linux-cachyos-rc/PKGBUILD
@@ -239,7 +239,10 @@ fi
 if [ "$_build_nvidia_open" = "yes" ]; then
     source+=("https://download.nvidia.com/XFree86/${_nv_open_pkg%"-$_nv_ver"}/${_nv_open_pkg}.tar.xz"
              "${_patchsource}/misc/nvidia/0002-Add-IBT-support.patch"
-             "${_patchsource}/misc/nvidia/0004-HACK-kernel-open-Makefile-Remove-PAHOLE_VARIABLE.patch")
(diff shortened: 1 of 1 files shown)


linux-cachyos-rc/PKGBUILD
```diff
@@ -237,9 +237,14 @@ fi
...
+    nvidia_patches=("0002-Add-IBT-support.patch"
+                    "0004-HACK-kernel-open-Makefile-Remove-PAHOLE_VARIABLE.patch"
+                    "0003-fix-dsc-correct-RC-parameter-tables-to-match-VESA-DS.patch"
+                    "0004-fix-dsc-use-bits_per_component-for-flatnessDetThresh.patch"
+                    "0005-fix-dp-add-Bigscreen-Beyond-VR-headset-to-WAR-databa.patch")
+
     source+=("https://download.nvidia.com/XFree86/${_nv_open_pkg%"-$_nv_ver"}/${_nv_open_pkg}.tar.xz"
-             "${_patchsource}/misc/nvidia/0002-Add-IBT-support.patch"
-             "${_patchsource}/misc/nvidia/0004-HACK-kernel-open-Makefile-Remove-PAHOLE_VARIABLE.patch")
+             "${nvidia_patches[@]/#/"${_patchsource}"/misc/nvidia/}")
```
1Naim: It's also possible to keep it in source and check for the nvidia prefix as in kernel patching. Let me know which is preferable.

1Naim:
Draft till the first commit is acceptable then I will port it to the other PKGBUILDs.

1Naim:
Will keep the DSC fixes only on rc for now.

PR #770, opened 2026-03-18
Fix configuration options in PKGBUILD

D3vil0p3r:


1Naim (commented):
These are not broken. scripts/config accepts config names with and without CONFIG_ prefixes.

PR #806, opened 2026-04-13
stable: 7.0.0-1

1Naim:
7.0 removed voluntary and none preemption so:
- kernels use full preemption and dynamic preemption is disabled by default
- server kernels will use lazy preemption
- rt kernels will have dynamic preemption enabled

Removed some unused variables

Handheld configs are now explicitly set in the PKGBUILD so configs from default kernel can finally be reused.


sirlucjan (approved):
LGTM

ptr1337 (commented):
Seems fine. I prepare default and deckify kernel for testing.

ptr1337:
## 7.0 Kernel Testing
### AMD/INTEL
```
sudo pacman -U https://archive.cachyos.org/kernel/7.0/default/linux-cachyos-7.0.0-0.1-x86_64_v3.pkg.tar.zst https://archive.cachyos.org/kernel/7.0/default/linux-cachyos-headers-7.0.0-0.1-x86_64_v3.pkg.tar.zst
```


### NVIDIA
```
sudo pacman -U https://archive.cachyos.org/kernel/7.0/default/linux-cachyos-7.0.0-0.1-x86_64_v3.pkg.tar.zst https://archive.cachyos.org/kernel/7.0/default/linux-cachyos-headers-7.0.0-0.1-x86_64_v3.pkg.tar.zst https://archive.cachyos.org/kernel/7.0/default/linux-cachyos-nvidia-open-7.0.0-0.1-x86_64_v3.pkg.tar.zst
```

ZFS and R8125 Module can be in the same path found

ventureoo:
> rt kernels will have dynamic preemption enabled

What is the point in this? `PREEMPT_RT` does not support dynamic preemption anyway.

1Naim:
> What is the point in this? `PREEMPT_RT` does not support dynamic preemption anyway

So users can switch between full and lazy there, since using lazy seems to bring back throughput quite a bit. See https://lore.kernel.org/all/579b7ea34ef6e2f7c955abdfc0929fe1af36faef.camel@gmx.de

While looking for that thread I also saw https://github.com/CachyOS/linux-cachyos/issues/339#issuecomment-4149360106 and yeah whatever we set there won't work. I guess we can remove the restriction, opinions?

ventureoo:
> So users can switch between full and lazy there, since using lazy seems to bring back throughput quite a bit. See https://lore.kernel.org/all/579b7ea34ef6e2f7c955abdfc0929fe1af36faef.camel@gmx.de

You mean building linux-cachyos-rt with _preempt=lazy? I don't see the point in that, actually I would like to move `PREEMPT_RT` out `_cpusched` and make it part of `_preempt` again.

1Naim:
> You mean building linux-cachyos-rt with _preempt=lazy

So they can set the preemption method at runtime instead of rebuilding.

> actually I would like to move `PREEMPT_RT` out `_cpusched` and make it part of `_preempt` again.

No opposition from my side, do we want it in 7.0 though, or defer it to 7.1 to not delay this?

ChiragM999:
Hey Folks, just curious here. Any ideas on when this will be released to master?

PR #842, opened 2026-05-09
linux-cachyos: Disable Clang as default since its broken

ptr1337:
The Clang Toolchain is currently not able to compile a kernel. This happens when compiling LLVM with GCC16.

This does not happen when LLVM is compiled with Clang or GCC15. There is currently no fast fix so we need to relay on gcc again


ventureoo:
I think this can be closed since issue was fixed?

CherrieTheShifter:
Yes This?

Issue #862, opened 2026-05-31
[FEATURE] Enable IMA so MOK-signed modules load under Secure Boot


mbocevski:
### Pre-flight checklist

- [x] I have searched existing issues/PRs and this has not been requested before.
- [x] I have read the Contributing Guidelines.

### Request type

- [x] Kernel config change (enable/disable option)
- [x] Security hardening

### Summary

`linux-cachyos` builds with IMA disabled (`# CONFIG_IMA is not set` and `# CONFIG_IMA_SECURE_AND_OR_TRUSTED_BOOT is not set`). With those off, `arch_ima_get_secureboot()` is a stub that always returns false (`include/linux/ima.h`), so `load_uefi_certs()` in `security/integrity/platform_certs/load_uefi.c` returns before `load_moklist_certs()` and MOK keys are never loaded into the `.machine` keyring:

```
/* the MOK/MOKx can not be trusted when secure boot is disabled */
if (!arch_ima_get_secureboot())
    return 0;
```

The result is that MOK-signed kernel modules can't be loaded under Secure Boot.

Proposed change: enable the same options the Fedora packaging already uses —
`CONFIG_IMA`, `CONFIG_IMA_APPRAISE`, `CONFIG_IMA_APPRAISE_BOOTPARAM`, `CONFIG_IMA_ARCH_POLICY`, `CONFIG_IMA_SECURE_AND_OR_TRUSTED_BOOT`. On EFI x86, `IMA_SECURE_AND_OR_TRUSTED_BOOT` is implied once IMA + IMA_ARCH_POLICY are on.

### Motivation / use case

With Secure Boot on (lockdown=integrity), DKMS modules (openrazer-dkms, nvidia-dkms, etc.) signed with the DKMS MOK key fail to load even after the key is enrolled and trusted:

```
modprobe: ERROR: could not insert 'razerkbd': Operation not permitted
Lockdown: modprobe: unsigned module loading is restricted
```

The key is enrolled and trusted (`mokutil --list-enrolled` and `mokutil --trust-mok` confirm it), but `.machine` stays empty. dmesg shows the `UEFI:db` certs loading into `.platform` but no `UEFI:MokListRT` line, even though MokListRT and MokListTrustedRT are present in the mokvar table. Enabling IMA fixes it.

[...]

PR #863, opened 2026-05-31
fix: enable IMA so MOK-signed modules load under Secure Boot

mbocevski:
Enables IMA across all kernel variants so MOK keys get loaded into the `.machine` keyring. Without it, `arch_ima_get_secureboot()` is a stub that returns false, `load_uefi_certs()` returns before `load_moklist_certs()`, and MOK-signed DKMS modules (e.g. openrazer) can't load under Secure Boot.

Same change already done for the Fedora packaging in copr-linux-cachyos (`a0407947`).

Scope of the behavior change: the arch IMA policy is only loaded when Secure Boot is enabled (`arch_get_ima_policy()` gates on `arch_ima_get_secureboot()`), so **non-Secure-Boot systems are unaffected**. Under Secure Boot it adds `measure` rules and calls `set_module_sig_enforced()`/`set_kexec_sig_enforced()` — i.e. it requires signed modules/kexec images. On a Secure Boot setup already running `lockdown=integrity` that is a no-op (lockdown enforces the same thing); on Secure Boot **without** lockdown it newly enforces signed modules. This kernel does not auto-enter lockdown under Secure Boot (`LOCK_DOWN_KERNEL_FORCE_NONE`), so that case is worth a heads-up.

Done as a `config` change in every variant (the resolved IMA block), matching how config options are flipped here (e.g. #836). `b2sums`/`.SRCINFO` still need regenerating with `srcinfo.sh`.

Tested on 7.0.10-2: `.machine` is populated and the previously-blocked modules load. Full writeup and logs in #862.

Docs: a companion wiki section (DKMS/MOK module signing under Secure Boot) is drafted in CachyOS/wiki#482, blocked on this PR.


CherrieTheShifter:
Yes This

PR #885, opened 2026-06-18
fix(cachyos-bore): update config file b2sum

Vikyek:
fix(cachyos-bore): update config file b2sum

The maintainer changed the config file settings in the latest commit
but forgot to update its corresponding hash in the b2sums array.
This caused the source file validation check to fail for all builds.

* Update config b2sum in PKGBUILD to the correct value
* Update config b2sum in .SRCINFO to match PKGBUILD


ptr1337:
Hey,

I would merge soon the 7.1 PR and then this should be fixed.

PR #938, opened 2026-07-22
linux-cachyos: Remove LLVM toolchain dependency on headers

### LINKED ISSUE #560: Drop the LLVM hard depends (linux-cachyos-headers)
As I understand it is only used for the kernel build and for the DKMS build. But for dkms the dependency on it is already specified. So it can be removed for those who do not use DKMS.

I may have made a mistake somewhere and misunderstood.

1Naim:
The headers itself don't depend on the LLVM toolchain, rather dkms requires it if the kernel is built with LLVM. That dependency is already handled there, so headers depending on LLVM is not needed.

Closes: #560


CherrieTheShifter:
Yes This

PR #976, opened 2026-08-13
add cachyos-infinity kernel

dim-geo:
[Infinity](https://github.com/galpt/infinity-scheduler) scheduler can be applied cleanly into cachyos.
Only patch 15 requires small modifications.
It builds and operates normally on my PC.


CherrieTheShifter:
Yes This Please:3
we need more performance esp cpu, definatly gpu and such
i hope they add this, would be peak tbh

also the [Infinity](https://github.com/galpt/infinity-scheduler) scheduler i think it updates, soooo

PR #1044, opened 2026-09-14
linux-cachyos, linux-cachyos-rc: fix drm/gud build break with clang 22 at -O3 (fixes #1031)

### LINKED ISSUE #1031: [BUG] the latest linux-cachyos-rc kernel failed to build
### Pre-flight checklist

- [x] I have searched existing issues and this is not a duplicate.
- [x] I have read the [Contributing Guidelines](https://github.com/CachyOS/linux-cachyos/blob/master/CONTRIBUTING.md).
- [x] I have verified the issue is reproducible with the latest available CachyOS kernel.
- [x] I have tried to reproduce the issue on Arch Linux's `linux` kernel.

### Upstream / vanilla kernel check

I have not tested with a vanilla/upstream kernel

### Kernel variant

linux-cachyos-rc

### System information (cachyos-bugreport.sh)

```text
Kernel version: 7.3.0-rc1-1-cachyos-rc
CPU: Intel Core Ultra X7 358H
Architecture: x86-64
```

### Manual system information (if cachyos-bugreport.sh is unavailable)

```text

```

### Bug description

The latest `linux-cachyos-rc` kernel fail to build with this `ld.lld` messages

```
ld.lld: error: call to __read_overflow marked "dontcall-error": detected read beyond size of object (1st parameter)
make[6]: *** [scripts/Makefile.build:510: drivers/gpu/drm/gud/gud.o] Error 1
make[5]: *** [scripts/Makefile.build:551: drivers/gpu/drm/gud] Error 2
make[4]: *** [scripts/Makefile.build:551: drivers/gpu/drm] Error 2
make[3]: *** [scripts/Makefile.build:551: drivers/gpu] Error 2
```

### Steps to reproduce

1. Build the kernel

### Expected behavior

The kernel will build successfully

### Actual behavior

Receive an error message of `ld.lld`

### Logs / stack traces

```text

```

### Additional system information

_No response_

mayankmohan1992:
## What this fixes

`linux-cachyos` and `linux-cachyos-rc` no longer build `drivers/gpu/drm/gud/` with clang 22 (issue #1031):

```
  CC [M]  drivers/gpu/drm/gud/gud_connector.o
  LD [M]  drivers/gpu/drm/gud/gud.o
ld.lld: error: call to __read_overflow marked "dontcall-error": detected read beyond size of object (1st parameter)
make[2]: *** [scripts/Makefile.build:550: drivers] Error 2
```

Both variants default to the combination that trips it (`_cc_harder=yes` → `KBUILD_CFLAGS -O3`, `_use_llvm_lto=thin`), and the offending code is in the 7.2.5-1 **and** 7.3-rc2 sources, so this blocks source/AUR builds of the current kernels. The other variants default to `_use_llvm_lto=none` (GCC), which is why they are not touched here — say the word and I will extend the patch to them as well, or move it to `kernel-patches/misc` and reference it via `${_patchsource}` (whichever the maintainers prefer for fix patches).

## Why it happens

`gud_connector_add_tv_mode()` computes `num_modes` from the length of the GUD reply and then walks both `modes[]` (`GUD_CONNECTOR_TV_MODE_MAX_NUM` = 16 entries) and a buffer of `MAX_NUM * NAME_LEN` = 256 bytes with that value:

```c
num_modes = ret / GUD_CONNECTOR_TV_MODE_NAME_LEN;
for (i = 0; i < num_modes; i++) {
        char *mode = &buf[i * GUD_CONNECTOR_TV_MODE_NAME_LEN];
        if (!memchr(mode, '\0', GUD_CONNECTOR_TV_MODE_NAME_LEN)) { ... }
```

`ret` is capped at `buf_len` by `gud_usb_get()` (which is handed `buf_len` as the transfer size), so `num_modes <= 16` always holds at runtime — but that invariant is invisible to the optimizer. Since upstream `da1ea35fea67` ("drm/gud: validate TV mode names before creating enum property") the fortified `memchr()` is a *provable* out-of-bounds read for LLVM, and the kernel's FORTIFY machinery (`__read_overflow` with `dontcall-error`) turns that into a hard build failure rather than a runtime check. It is reported by the front end when LTO is off:

```
include/linux/fortify-string.h:686:3: error: call to '__read_overflow' declared with 'error' attribute: detected read beyond size of object (1st parameter)
```

## The patch

`0001-drm-gud-bound-tv-mode-loop.patch` (applied by `prepare()` automatically, same mechanism as the other `source=()` patches):

* reject replies that claim more modes than the protocol defines — defensive against a misbehaving USB device, and it keeps `drm_mode_create_tv_properties_legacy()` from walking `modes[]` entries that were never initialised;
* make the loop bound explicit so the compiler can prove the accesses are in range.

Both halves are required. Measured on this machine (extracted 7.2.5-1 sources, `linux-cachyos/config` plus the PKGBUILD's default knobs, `make LLVM=1 LLVM_IAS=1 -j12 drivers/gpu/drm/gud/`):

| tree | result |
|---|---|
| unpatched | **exit 2** — `ld.lld: error: call to __read_overflow ...` |
| reject-only (check, no loop bound) | **exit 2** — the bound is what the optimizer needs |
| reject + loop bound (this patch) | **exit 0** — `LD [M] drivers/gpu/drm/gud/gud.o` links |
| same patch, GCC / `LTO=none` | **exit 0** — no regression |

Effective config used for the clang run: `CONFIG_CC_IS_CLANG=y`, `CONFIG_CC_OPTIMIZE_FOR_PERFORMANCE_O3=y`, `CONFIG_LTO_CLANG_THIN=y`, `CONFIG_FORTIFY_SOURCE=y`, `CONFIG_DRM_GUD=m`, `CONFIG_X86_NATIVE_CPU=y`, clang/ld.lld 22.1.8.

### Bounds checked

* `patch -Np1` applies cleanly to the 7.2.5-1 tree **and** to the 7.3-rc2 tree (identical hunk context), so the same file is valid for both variants.
* Re-applying it to an already-patched tree is a no-op: `patch -Np1` prints `Reversed (or previously applied) patch detected! Skipping patch.` and exits **0** — so the build still succeeds if this lands upstream first.
* `b2sums=()` is extended in the same position as `source=()`. This matters: makepkg hard-fails with `ERROR: Integrity checks (b2) differ in size from the source array.` when the two arrays differ (verified with a minimal PKGBUILD), so adding the source line alone would have broken every AUR build. `makepkg --verifysource` now reports 5 sources against 5 checksums with `0001-drm-gud-bound-tv-mode-loop.patch ... Passed` for both variants.
* The patch touches a single function; the only behaviour change is that a reply longer than the protocol's maximum is rejected with `-EIO` (previously it could not happen, and if it somehow did, `modes[]` would have been overrun). It should be dropped once an equivalent bound lands upstream.

Fixes #1031


linux-cachyos/0001-drm-gud-bound-tv-mode-loop.patch
```diff
@@ -0,0 +1,55 @@
...
+-	for (i = 0; i < num_modes; i++) {
++	if (num_modes > GUD_CONNECTOR_TV_MODE_MAX_NUM) {
++		ret = -EIO;
++		goto free;
++	}
++
++	for (i = 0; i < num_modes && i < GUD_CONNECTOR_TV_MODE_MAX_NUM; i++) {
+ 		char *mode = &buf[i * GUD_CONNECTOR_TV_MODE_NAME_LEN];
+ 
+ 		if (!memchr(mode, '\0', GUD_CONNECTOR_TV_MODE_NAME_LEN)) {
```
n0099, line 55: Your patches are same with the one in https://github.com/MastaG/linux-cachyos-bc250/commit/db49878af40551b481f511053201fcf1e1bd5d90#diff-7a8c6d5b1b5e42b00007c96773ce0a1673b1f0ae16045d0be50212a8862ba07bR36-R56 that created by @MastaG's LLM: https://github.com/CachyOS/linux-cachyos/issues/1031#issuecomment-5664110094

vnepogodin:
AI slop maxxing

hotline1337 (commented):
Naim already fixed this with commit [d10310a](https://github.com/CachyOS/linux/commit/d10310a163cefdadf7e42689243b15eef85a5435) on `7.3/fixes` branch. This PR can be closed.

ventureoo:
Superseded by https://github.com/CachyOS/linux/commit/d10310a163cefdadf7e42689243b15eef85a5435

